PR #3, opened 2020-12-07
Add proper backward compatible cases when unmarshaling authType

aocenas:
Seems like this is the correct mapping for older option names but if any data source is already using this they may rely on the older mapping in some way. But I assume only other one that needs to be checked besides x-ray is timescale.


ryantxu (approved):
thanks!

ryantxu:
I'm going to merge so I can update in sitewise & timestream

PR #4, opened 2020-12-11
Add SigV4 middleware

wbrowne:
Fixes https://github.com/grafana/grafana/issues/29738

Usage looks like:

```go
return sigv4.New(
    &Config{
        Service:        "es",
        AccessKey:      "blah",
        SecretKey:      "blah",
        Region:         "eu-central-1",
        AssumeRoleARN:  "", // optional
        AuthType:       "default",
        ExternalID:     "", // optional
        Profile:        "default",
   },
   nil, // optional subsequent middleware
)
```

### DIFF
--- a/pkg/sigv4/sigv4.go
+++ b/pkg/sigv4/sigv4.go
@@ -1,0 +1,161 @@
+package models
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/session"
+	v4 "github.com/aws/aws-sdk-go/aws/signer/v4"
+	"github.com/aws/aws-sdk-go/private/protocol/rest"
+)
+
+type AuthType string
+
+const (
+	Default     AuthType = "default"
+	Keys        AuthType = "keys"
+	Credentials AuthType = "credentials"
+)
+
+// Host header is likely not necessary here
+// (see https://github.com/golang/go/blob/cad6d1fef5147d31e94ee83934c8609d3ad150b7/src/net/http/request.go#L92)
+// but adding for completeness
+var permittedHeaders = map[string]struct{}{
+	"Host":            {},
+	"Uber-Trace-Id":   {},
+	"User-Agent":      {},
+	"Accept":          {},
+	"Accept-Encoding": {},
+}
+
+type SigV4Middleware struct {
+	Config *Config
+	Next   http.RoundTripper
+}
+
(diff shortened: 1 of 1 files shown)


wbrowne:
Will need to keep in sync with these latest changes here https://github.com/grafana/grafana/pull/30115

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,150 @@
...
+	"Accept":          {},
+	"Accept-Encoding": {},
+}
+
+type SigV4Middleware struct {
+	Config *Config
+	Next   http.RoundTripper
+}
+
+type Config struct {
```
marefr, line 43: Nit. Don't need to handle in this PR, but moving forward think this is very similar to awsds auth and would be nice if the actual credentials creation code was shared, but not sure that's possible?

marefr (changes requested):
Some minor things, see comments. Comments around sharing the credentials creation/models in this lib is something to consider/discuss later.

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/session"
+	v4 "github.com/aws/aws-sdk-go/aws/signer/v4"
+	"github.com/aws/aws-sdk-go/private/protocol/rest"
+)
+
+type AuthType string
```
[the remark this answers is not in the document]
wbrowne: Yep - agreed!

pkg/sigv4/sigv4.go
wbrowne: I think it should be. Will confer with @ryantxu as a form of refactoring afterwards

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,170 @@
...
+	return rt(r)
+}
+
+// New instantiates a new signing middleware with an optional succeeding
+// middleware. The http.DefaultTransport will be used if nil
+func New(config *Config, next http.RoundTripper) http.RoundTripper {
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		if next == nil {
+			next = http.DefaultTransport
+		}
```
marefr, line 74: Nit. Would rather require the client to provide `next` to make it more explicit that the middleware wont function without `next` provided. But that would either require an error to be returned from `New` or potentially a `panic` but that's ugly.

marefr (approved):
LGTM, but added a comment regarding the optional middleware

pkg/sigv4/sigv4.go
wbrowne: Yeah, that's a potential scenario. I think let's keep it as is and can refactor that out (or just add another alternative `New` func) later if necessary

PR #5, opened 2021-02-18
Prepare npm package and update shared component

sunker:
**Background**
Currently there are five different AWS plugins and there are more to come this year. Four of the plugins are external, and one of them is in core grafana (CloudWatch). All the plugins have basically the same needs when it comes to handling authentication, so if this code could be shared it would save us a lot of time when changes to auth providers are made. There is already a go module that makes in possible to share backend code, but we also need an npm package that enables sharing frontend types and a base ConfigEditor component. 

**What this PR does** 
* update rollup bundle and npm package configuration
* refactor shared ui component

The name of the package is @grafana/aws-sdk. Currently it's published to npm and our github package registry. The lastest version is 0.0.153. If you have problems consuming packages from our github package registry, refer to [this](https://github.com/grafana/grafana/blob/master/packages/README.md#consuming-prereleases) guide. 

Instruction on how to enable/disable auth providers and enable/disable assume role can be found in the aws section in the [config](https://github.com/grafana/grafana/blob/master/docs/sources/administration/configuration.md#L775).  

Don't have that much knowledge around aws plugins except for cloudwatch, so looking to get some feedback on the api.  

Currently I have a [draft PR](https://github.com/grafana/grafana/pull/31807) open in the grafana repo that aims to consume this package.

**Question**
Do we want the go packages and the npm package to share version number? In that case we should add CI/CD for this repo.


src/types.ts
```diff
@@ -1,9 +1,14 @@
...
   Credentials = 'credentials',
   Default = 'default', // was 'arn',
+  EC2IAMRole = 'ec2_IAM_role',
+  /**
+   * @deprecated use default
+   */
+  ARN = 'arn',
 }
 
 export interface AwsDataSourceJsonData extends DataSourceJsonData {
```
aocenas, line 14: I think this should be named something like `AwsAuthDataSourceJsonData` just to be clear this is not all the data needed for the data source and you should be extending it.

src/ConnectionConfig.tsx
```diff
@@ -1,238 +1,158 @@
...
+      <h3 className="page-heading">Connection Details</h3>
+      <InlineField
+        label="Authentication Provider"
+        labelWidth={28}
+        tooltip="Specify which AWS credentials chain to use. AWS SDK Default is the recommended option for EKS, ECS, or if you've attached an IAM role to your EC2 instance."
+      >
+        <Select
+          className="width-30"
+          value={awsAuthProviderOptions.find((p) => p.value === options.jsonData.authType) || awsAuthProviderOptions[0]}
+          options={awsAuthProviderOptions.filter((opt) => config.awsAllowedAuthProviders.includes(opt.value!))}
```
aocenas, line 50: What in cases there already are data sources with option that we will make unavailable after the config change? Maybe we should add option that is already selected so it doesn't just disappear but add some sort of warning about it?

src/types.ts
sunker: I agree!

src/ConnectionConfig.tsx
sunker: For OSS Grafana, options will be the same as before. AMG is not officially released yet, so that shouldn't be a problem. For Hosted grafana on the other hand, that might be a problem. I know they want to be able to disable `Credentials file` since it's not possible to upload a file to server in HG anyway.

package.json
```diff
@@ -1,23 +1,31 @@
...
   "license": "Apache-2.0",
   "dependencies": {
-    "@grafana/data": "latest",
-    "@grafana/ui": "latest"
+    "@grafana/data": "7.5.0-beta.1",
+    "@grafana/runtime": "7.5.0-beta.1",
+    "@grafana/ui": "7.5.0-beta.1"
   },
   "devDependencies": {
     "@grafana/toolkit": "latest",
```
dprokop, line 27: This is my only concern- if you need toolkit in this package, since it's only used for configs (prettier, tsconfig).

The toolkit's pluging tsconfig is fairly simple, and I think you could use @grafana/tsconfig instead of importing the config from the toolkit.

We don't have a shared prettier config tho, but internally we also use toolkit's one: https://github.com/grafana/grafana/blob/017767ae39a10bedb62c82a264a351ab5357bb77/packages/grafana-toolkit/src/config/prettier.plugin.config.json#L5

Maybe creating @grafana/prettier-config package is a way to go, to remove get rid of the prettier imports from toolkit.

dprokop:
Looks good config-wise @sunker. Left one comment about toolkit dependency that I believe can be remove.

package.json
dprokop: Sounds good!

PR #8, opened 2021-03-10
Align auth code with cloudwatch

sunker:


pkg/awsds/sessions.go
```diff
@@ -52,15 +55,36 @@ var newSTSCredentials = stscreds.NewCredentials
...
+	}
+
+	if s.AssumeRoleARN != "" && !sc.authSettings.AssumeRoleEnabled {
+		return nil, fmt.Errorf("attempting to use assume role (ARN) which is disabled in grafana.ini")
+	}
+
 	bldr := strings.Builder{}
 	for i, s := range []string{
-		s.AuthType.String(), s.AccessKey, s.Profile, s.AssumeRoleARN, region,
+		s.AuthType.String(), s.AccessKey, s.Profile, s.AssumeRoleARN, region, s.Endpoint,
```
ryantxu, line 87: interesting -- in sitewise and timestream, we pass the endpoint as part of the client, not the session:

https://github.com/grafana/timestream-datasource/blob/master/pkg/timestream/datasource.go#L71
https://github.com/grafana/iot-sitewise-datasource/blob/9e0dd3d03d5864a9a784c01d8afe42b222d03063/pkg/sitewise/client/client.go#L100

I know that iot-sitewise needs the endpoint configured on the client, not the session --- but I don't know if it is bad to do it in both or just redundant

pkg/awsds/sessions.go
```diff
@@ -26,12 +27,14 @@ type envelope struct {
 type SessionCache struct {
 	sessCache     map[string]envelope
 	sessCacheLock sync.RWMutex
+	authSettings  *AuthSettings
 }
 
 // NewSessionCache creates a new session cache
-func NewSessionCache() SessionCache {
+func NewSessionCache(authConfig *AuthSettings) SessionCache {
```
ryantxu, line 34: Can we make sure plugins do not need to create this settings object themselves?  We need a single strategy that will work for all AWS plugins and should avoid needing to update too much code themselfs

pkg/awsds/sessions.go
```diff
@@ -26,12 +29,42 @@ type envelope struct {
...
+	for _, authProvider := range strings.Split(providers, ",") {
+		authProvider = strings.TrimSpace(authProvider)
+		if authProvider != "" {
+			allowedAuthProviders = append(allowedAuthProviders, authProvider)
+		}
+	}
+
+	if len(allowedAuthProviders) == 0 {
+		allowedAuthProviders = []string{"default", "keys", "credentials"}
+		plog.Warn("could not find allowed auth providers. falling back to 'default, keys, credentials'")
```
sunker, line 56: this is not ideal...I'd like to return an error, but that would break the existing api of creating sessions.

ryantxu: if we really want things to fail hard if we are missing expected environment variables, then we should create a common function and call it in the startup method of the various aws methods

Essentially add a one line function here:
https://github.com/grafana/iot-sitewise-datasource/blob/main/pkg/main.go#L13

pkg/awsds/sessions.go
```diff
@@ -26,12 +29,50 @@ type envelope struct {
...
+func NewSessionCacheWithSettigns(s *AuthSettings) SessionCache {
 	return SessionCache{
-		sessCache: map[string]envelope{},
+		sessCache:    map[string]envelope{},
+		authSettings: s,
+	}
+}
+
+const ENV_VAR_AllowedAuthProviders = "AWS_AUTH_AllowedAuthProviders"
+const ENV_VAR_AssumeRoleEnabled = "AWS_AUTH_AssumeRoleEnabled"
```
ryantxu, line 50: TODO: pick reasonable environment variable names/pattern/style

pkg/awsds/sessions.go
```diff
@@ -118,7 +187,7 @@ func (sc *SessionCache) GetSession(region string, s AWSDatasourceSettings) (*ses
 
 	duration := stscreds.DefaultDuration
 	expiration := time.Now().UTC().Add(duration)
-	if s.AssumeRoleARN != "" {
+	if s.AssumeRoleARN != "" && sc.authSettings.AssumeRoleEnabled {
 		// We should assume a role in AWS
 		plog.Debug("Trying to assume role in AWS", "arn", s.AssumeRoleARN)
```
sunker, line 193: I've taken a closer look at how we handle region in this method. We should only ever use region from AWSDatasourceSettings struct in case the free region is 'default'. That's the whole idea of the of the default region config in the data source configuration page.

pkg/awsds/sessions_test.go
```diff
@@ -57,7 +58,7 @@ func TestNewSession_AssumeRole(t *testing.T) {
 			AssumeRoleARN: roleARN,
 		}
 
-		cache := NewSessionCache()
+		cache := NewSessionCacheWithSettigns(&AuthSettings{AllowedAuthProviders: []string{"default"}, AssumeRoleEnabled: true})
```
aknuds1, line 61: Not sure where this function is defined, but it has a typo.
```suggestion
		cache := NewSessionCacheWithSettings(&AuthSettings{AllowedAuthProviders: []string{"default"}, AssumeRoleEnabled: true})
```

aknuds1 (approved):
LGTM, except would be great if you could fix the typo I point out.

pkg/awsds/sessions_test.go
sunker: Oh yes, I'll fix!

ryantxu (approved):
nice -- also note that the changed method signature (now returning a pointer) did not effect the sitewise plugin code.

````

	sessions := awsds.NewSessionCache()

	return &Datasource{
		GetClient: func(region string) (swclient client.SitewiseClient, err error) {
			swclient, err = client.GetClient(region, cfg, sessions.GetSession)
			return
		},
	}, nil
```

although I suppose it could panic if that were a nil

PR #9, opened 2021-03-17
Fix: ConnectionConfig issues

zoltanbedi:
This pr fixes the issues Andrej found during review https://github.com/grafana/x-ray-datasource/pull/73#issuecomment-800245277

### DIFF
--- a/src/ConnectionConfig.tsx
+++ b/src/ConnectionConfig.tsx
@@ -49,7 +49,6 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
 
   const currentProvider = awsAuthProviderOptions.find((p) => p.value === options.jsonData.authType);
 
-  // Component did mount
   useEffect(() => {
     // Make sure a authType exists in the current model
     if (!currentProvider && awsAllowedAuthProviders.length) {
--- a/src/ConnectionConfig.tsx
+++ b/src/ConnectionConfig.tsx
@@ -18,6 +18,13 @@ import { AwsAuthType } from 'types';
 
 const toOption = (value: string) => ({ value, label: value });
 
+// awsAllowedAuthProviders is supported in 7.5+
+const awsAllowedAuthProviders: AwsAuthType[] = (config as any).awsAllowedAuthProviders ?? [
+  AwsAuthType.Default,
+  AwsAuthType.Keys,
+  AwsAuthType.Credentials,
+];
+
 export interface ConnectionConfigProps<J = AwsAuthDataSourceJsonData, S = AwsAuthDataSourceSecureJsonData>
   extends DataSourcePluginOptionsEditorProps<J, S> {
   standardRegions?: string[];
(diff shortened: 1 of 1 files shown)


src/ConnectionConfig.tsx
```diff
@@ -107,6 +108,8 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
                 <ToolbarButton
                   icon="edit"
                   tooltip="Edit Access Key ID"
+                  // this can be rewritten to type="button" after upgrading to @grafana/ui@7.5
```
aocenas, line 111: Why cannot it be written that way now? The `{...{ type: 'button' }}` basically does the same think doesn't it?

src/ConnectionConfig.tsx
```diff
@@ -36,34 +43,28 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
...
     props.loadRegions().then((regions) => setRegions(regions.map(toOption)));
-  }, [props.loadRegions]);
-
-  // awsAllowedAuthProviders is supported in 7.5+
-  const awsAllowedAuthProviders: Array<AwsAuthType> = (config as any).awsAllowedAuthProviders ?? [
-    AwsAuthType.Default,
-    AwsAuthType.Keys,
-    AwsAuthType.Credentials,
-  ];
+  }, [props]);
```
aocenas, line 67: Can we change this to only use `loadRegions`? Otherwise it seems like this change will make this run on any props change which isn't how it worked before.

src/ConnectionConfig.tsx
```diff
@@ -36,34 +43,28 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
   }
 
   const currentProvider = awsAuthProviderOptions.find((p) => p.value === options.jsonData.authType);
+
   // Component did mount
   useEffect(() => {
     // Make sure a authType exists in the current model
-    if (!currentProvider && awsAuthProviderOptions.length) {
+    if (!currentProvider) {
```
aocenas, line 50: Isn't it possible for `awsAuthProviderOptions.length` to be 0? I assume that could happen if `(config as any).awsAllowedAuthProviders === []` or something like that? Does that make sense to configure it like that?

src/ConnectionConfig.tsx
zoltanbedi: Typescript has a problem with it right now, it says that there isn't a type attribute in the component.

src/ConnectionConfig.tsx
zoltanbedi: ohh you are right, that doesn't make sense.

src/ConnectionConfig.tsx
```diff
@@ -28,42 +40,36 @@ export interface ConnectionConfigProps<J = AwsAuthDataSourceJsonData, S = AwsAut
...
+  const { loadRegions, onOptionsChange } = props;
   const options = props.options;
   let profile = options.jsonData.profile;
   if (profile === undefined) {
     profile = options.database;
   }
 
   const currentProvider = awsAuthProviderOptions.find((p) => p.value === options.jsonData.authType);
+
   // Component did mount
```
sunker, line 52: Not true anymore so this comment can be removed

src/ConnectionConfig.tsx
zoltanbedi: This fixes what Andrej found that when someone presses enter in the input field (for example in the data source name) it triggers onSubmit which triggers all the submit type buttons click as well. The button type is submit by default.

sunker: I see

PR #12, opened 2021-03-26
Add support for Workspace IAM with SigV4

wbrowne:


pkg/sigv4/sigv4.go
```diff
@@ -109,15 +108,37 @@ func (m *middleware) signRequest(req *http.Request) (http.Header, error) {
...
+	for _, provider := range allowedAuthProviders {
+		if provider == authType.String() {
+			authTypeAllowed = true
+			break
+		}
+	}
+
+	if !authTypeAllowed {
+		return nil, fmt.Errorf("attempting to use an auth type that is not allowed: %q", authType.String())
+	}
```
sunker, line 124: Can we also check if assume role was used and if it's allowed? The same way we do in awsds.GetSession

pkg/sigv4/sigv4.go
```diff
@@ -109,15 +108,37 @@ func (m *middleware) signRequest(req *http.Request) (http.Header, error) {
...
-	case Keys:
+	case awsds.AuthTypeKeys:
 		c = credentials.NewStaticCredentials(m.config.AccessKey, m.config.SecretKey, "")
-	case Credentials:
+	case awsds.AuthTypeSharedCreds:
 		c = credentials.NewSharedCredentials("", m.config.Profile)
-	case Default:
+	case awsds.AuthTypeEC2IAMRole:
+		s, err := session.NewSession(&aws.Config{
+			Region: aws.String(m.config.Region),
```
sunker, line 133: Add endpoint

pkg/sigv4/sigv4.go
```diff
@@ -109,15 +108,37 @@ func (m *middleware) signRequest(req *http.Request) (http.Header, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		c = credentials.NewCredentials(&ec2rolecreds.EC2RoleProvider{Client: ec2metadata.New(s), ExpiryWindow: stscreds.DefaultDuration})
+
+		return v4.NewSigner(c), nil
+	case awsds.AuthTypeDefault:
 		// passing nil credentials will force AWS to allow a more complete credential chain vs the explicit default
 		s, err := session.NewSession(&aws.Config{
 			Region: aws.String(m.config.Region),
```
sunker, line 144: Add endpoint

pkg/sigv4/sigv4.go
```diff
@@ -131,7 +152,7 @@ func (m *middleware) signer() (*v4.Signer, error) {
 		}
 
 		return v4.NewSigner(s.Config.Credentials), nil
-	case "":
+	default:
 		return nil, fmt.Errorf("invalid SigV4 auth type")
 	}
```
sunker, line 158: Also check here if assume role is enabled

sunker (commented):
Looks like a great start! 

It sounds like aws wants the web form for sigv4 to be exactly the same as for cloudwatch (and the other plugins that use awsds), and therefore we need to add `Endpoint`. Read more about endpoint [here](https://grafana.com/docs/grafana/latest/datasources/cloudwatch/#endpoint). The way to specify an endpoint in the backend is to add it to the `aws.Config` whenever a new session in created in the signer function. 

I think that we could share more code between awsds and sigv4. For example, they should be using the same code to resolve credentials for each auth provider, and also they should be using the same code to check if an auth provider is allowed etc. However, that will need some refactoring in the awsds package and I'm not too excited about making changes there since it's already been tested by us, and AWS are currently testing it.

sunker (approved):
Great work!

PR #14, opened 2021-03-30
Chore: use dev dependencies

ryantxu:
The runtime environment for `grafana/ui` and `grafana/data` should come from the environment, not the aws bundle.

Making this change will likely avoid having the explicitly versioned grafana/ui&data in the cloudwatch bundle.  See:
https://github.com/grafana/grafana/issues/32328


ryantxu:
close in favor of #15

PR #15, opened 2021-03-30
Remove dependency to runtime. Read conf from window.grafanaBootData instead

sunker:


src/types.ts
```diff
@@ -24,3 +24,14 @@ export interface AwsAuthDataSourceSecureJsonData {
   accessKey?: string;
   secretKey?: string;
 }
+
+declare global {
```
ryantxu, line 28: I assume this *only* affects the typing?  it has no runtime effects?

ryantxu (approved):
LGTM

src/types.ts
sunker: reverting to use any instead

PR #19, opened 2021-05-10
Add Content-Length to header allowlist

wbrowne:
Without this, there will likely be verification issues when using POST


marefr (approved):
LGTM

Wasn't this removed a while ago or something? :)

PR #20, opened 2021-06-18
feat: avoid recreating the signer at every request sign

vthiery:
Looking into the code, I noticed that the `signer` is recreated every time we sign a request.
Was there a specific reason to do so?

I changed the code so that it is only created once, the benchmark I added and measured the following:

```
BenchmarkSignRequest-old   	   18208	     59109 ns/op	   12944 B/op	     124 allocs/op
BenchmarkSignRequest-new   	   72361	     17252 ns/op	    5960 B/op	      83 allocs/op
```

I also added a few unit tests (it is not straightforward to test the rest because of the global `authSettings` variable).

### DIFF
--- a/pkg/sigv4/sigv4.go
+++ b/pkg/sigv4/sigv4.go
@@ -2,6 +2,7 @@ package sigv4
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -42,6 +43,8 @@ var (
 type middleware struct {
 	config *Config
 	next   http.RoundTripper
+
+	signer *v4.Signer
 }
 
 type Config struct {
@@ -98,9 +101,12 @@ func (m *middleware) exec(req *http.Request) (*http.Response, error) {
 }
 
 func (m *middleware) signRequest(req *http.Request) (http.Header, error) {
-	signer, err := m.signer()
-	if err != nil {
-		return nil, err
+	if m.signer == nil {
+		signer, err := newSigner(m.config)
+		if err != nil {
+			return nil, err
+		}
+		m.signer = signer
 	}
 
 	body, err := replaceBody(req)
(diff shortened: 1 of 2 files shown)


wbrowne:
Thanks for the contribution @vthiery - will try get around to this ASAP!

sunker:
@wbrowne do you have an update on this one? 

I wonder if this should be fixed by the @grafana/observability-squad since they own all plugins that make use of SIGV4? Currently, our team has no way of testing this.

wbrowne:
@sunker No updates from my side as I haven't had a chance to invest any time in this.

Yeah I think that makes the most sense now. @grafana/observability-squad let me know if you need a hand with taking this over

aocenas:
@wbrowne @sunker so I feel like there may be some confusion here :D. I assume this is used in every AWS plugin right? We maintain CloudWatch logs and x-ray but to be honest I have no idea how to test the SIGV4 part and nobody from Observability was part of the SDK development.

wbrowne:
Also Prometheus and Elasticsearch @aocenas. I chatted about this in amongst the plugins platform group today and have an idea about who might be best to take this over. Will follow up internally

sunker:
Yes Prometheus and Elasticsearch are the only data sources that use SIGV4 auth.

gabor:
@wbrowne for now it seems the Observability Squads cannot help here, so i will un-assign the observability-project.
if this changes, we'll gladly help (there are now separate observability squads for metrics & logs-traces, so Prometheus belongs to @grafana/observability-metrics and ElasticSearch to @grafana/observability-logs-and-traces )

wbrowne:
I've just merged a [PR](https://github.com/grafana/grafana-aws-sdk/pull/48) that reworks our approach to SigV4 and will cache signers - so therefore I will close this one. Thanks for highlighting this @vthiery and apologies for the delay in getting it patched

PR #21, opened 2021-06-24
[SigV4] Add assume role support when using EC2 IAM auth

wbrowne:


pkg/sigv4/sigv4.go
```diff
@@ -151,6 +151,18 @@ func (m *middleware) signer() (*v4.Signer, error) {
 		}
 		c = credentials.NewCredentials(&ec2rolecreds.EC2RoleProvider{Client: ec2metadata.New(s), ExpiryWindow: stscreds.DefaultDuration})
 
+		if m.config.AssumeRoleARN != "" {
+			s, err = session.NewSession(&aws.Config{
+				CredentialsChainVerboseErrors: aws.Bool(true),
```
sunker, line 156: Do we want to keep this?

sunker (approved):
LGTM! 

I guess we can remove line 193-202 now?

pkg/sigv4/sigv4.go
wbrowne: Yeah I believe so. This will help give more useful error messages. It's something we already do for sessions.

wbrowne:
> LGTM!
> 
> I guess we can remove line 193-202 now?

Yeah these lines are still relevant for the `AuthTypeKeys` and `AuthTypeSharedCreds` cases where we only set the credentials

sunker:
> > LGTM!
> > I guess we can remove line 193-202 now?
> 
> Yeah these lines are still relevant for the `AuthTypeKeys` and `AuthTypeSharedCreds` cases where we only set the credentials

Okay cool!

wbrowne:
Tested and looks good!

PR #23, opened 2021-07-23
include session token in key creds

atifali:
This includes a param for adding a `sessionToken` value while authenticating with AWS using key pairs.


ryantxu (approved):
LGTM

macrovve:
Should we also change the UI, so that customer could enter their session token @atifali @ryantxu

PR #24, opened 2021-07-28
feature: support temporary security credentials

atifali:
this includes support for using temporary key credentials for AWS resources. exposes the `sessionToken` param, which is needed for this case.

see:
https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#temporary-access-keys
and:
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

### DIFF
--- a/src/ConnectionConfig.tsx
+++ b/src/ConnectionConfig.tsx
@@ -137,6 +137,26 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
               />
             )}
           </InlineField>
+
+          <InlineField label="Session Token" labelWidth={28}>
+            {props.options.secureJsonFields?.sessionToken ? (
+              <ButtonGroup className="width-30">
+                <Input disabled placeholder="Configured" />
+                <ToolbarButton
+                  icon="edit"
+                  type="button"
+                  tooltip="Edit Session Token"
+                  onClick={onUpdateDatasourceResetOption(props as any, 'sessionToken')}
+                />
+              </ButtonGroup>
+            ) : (
+              <Input
+                className="width-30"
+                value={options.secureJsonData?.sessionToken ?? ''}
+                onChange={onUpdateDatasourceSecureJsonDataOption(props, 'sessionToken')}
+              />
+            )}
+          </InlineField>
         </>
       )}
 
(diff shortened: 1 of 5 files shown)


src/ConnectionConfig.tsx
```diff
@@ -137,6 +137,26 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
...
+          <InlineField label="Session Token" labelWidth={28}>
+            {props.options.secureJsonFields?.sessionToken ? (
+              <ButtonGroup className="width-30">
+                <Input disabled placeholder="Configured" />
+                <ToolbarButton
+                  icon="edit"
+                  type="button"
+                  tooltip="Edit Session Token"
+                  onClick={onUpdateDatasourceResetOption(props as any, 'sessionToken')}
+                />
```
ryantxu, line 150: :thinking: The bad part here is that if we include this option and someone sets the value.... there is no way to clear it!!!!

https://github.com/grafana/grafana/issues/22552

sunker:
Thanks for contributing @atifali! I think this sounds like a nice addition to the Keys and secret auth, but let me run this through our contact person at AWS before moving forward with this.

atifali:
@sunker any word on this yet?

sunker:
I'll have an update later this week @atifali

src/ConnectionConfig.tsx
ryantxu: @atifali , can you remove this from the PR?  Supporting session token in the models is useful on its own, and we can figure out a solution here later

sunker:
> I'll have an update later this week @atifali

Still waiting for feedback from AWS in this matter. Hopefully they'll get back to me in the next two weeks.

ryantxu (approved):
@sunker -- lets keep this out of the UI for now.  since we can not clear the tokens it will for sure cause issues!

sunker:
I was holding my review until I got confirmation from the AWS folks. 

This PR only adds support to SIGV4 auth. It assume it needs to be added also to the normal auth flow? @atifali can you please open a new PR and add the session token args to [this](https://github.com/grafana/grafana-aws-sdk/blob/e6622ac3d426585289b910fd25e5456c475684b1/pkg/awsds/sessions.go#L176) line of code?

Issue #27, opened 2021-10-21
Setup CI/CD for this repo


sunker:
Currently, the only way to publish to npm and our github registry is by running 
```
yarn build
yarn bundle 
npm publish  --access=public --scope=@grafana
```
This requires you to be logged into npm and to be part of the grafana org. Not sure exactly how, but I'd like these steps to be automated. See [this](https://github.com/grafana/eslint-config-grafana/pull/10) PR to get an idea of how this has been solved in other repos at grafana.

For the backend, we could possibly have a bot create new releases every time we merge to the main branch. This is still TBD. 

This should be done after this repo is split up in two. See [this](https://github.com/grafana/grafana-aws-sdk/issues/26) issue.

andresmgot:
This was actually done with #30  and https://github.com/grafana/grafana-aws-sdk-react/pull/1

PR #30, opened 2021-10-29
Add basic drone config

andresmgot:
First step at #27 

This needs to add drone but I still don't have admin permissions.


sunker (approved):
LGTM

Issue #28, opened 2021-10-29
Common AWS SQL Datasources


andresmgot:
See design document at: https://docs.google.com/document/d/1V3SJGER5zIEQ_x-jHTE3gUIOvDMYqeYFsG4jxtTdux4/edit#

The objective of the refactoring is to reduce the amount of time spent on maintaining and developing new and existing SQL data sources for AWS, having a bigger source of common code.

Chores:
- [x] https://github.com/grafana/grafana-aws-sdk/issues/27 
- [x] https://github.com/grafana/grafana-aws-sdk/issues/26 

Backend
- [x] Create an API package with a common interface
- [x] Create a Driver package with a basic interface
- [x] Create a “routes” package with common routes
- [x] Create a datasource package for handling connections
- [x] Expose method to set the User-Agent header https://github.com/grafana/grafana-aws-sdk/pull/36

Frontend
- [x] ResourceSelector component
- [x] ConfigEditor component
- [x] QueryCodeEditor component
- [x] QueryEditor component (with support for examples)
- [x] Suggestions utils
~~- [ ] VariableSupport class~~
- [x] Datasource class

Athena
- [x] Migrate plugin to use the common packages

Redshift
- [x] Migrate plugin to use the common packages

Timestream
~~- [ ] Migrate plugin to use sqlds (requires a golang SQL driver)~~
~~- [ ] Migrate plugin to use the common packages~~

andresmgot:
Created a different issue/epic for Timestream: https://github.com/grafana/timestream-datasource/issues/124 

So we can consider this done.

PR #32, opened 2021-11-02
Add pkg/sql/routes package

andresmgot:
Package for AWS SQL data sources that define common routes.

Epic at #28
Design doc

This PR adds a `routes` package that defines routes to get `/regions` and `/databases`, using the API methods defined in #29 to get them.


pkg/sql/routes/routest_test.go
```diff
@@ -0,0 +1,122 @@
+package routes
```
sunker, line 1: filename typo

sunker (approved):
LGTM

pkg/sql/routes/routest_test.go
andresmgot: :joy: but it's a nice typo

PR #34, opened 2021-11-03
Add the rest of backend sql packages

### LINKED ISSUE #31: Common AWS SQL Datasource - Backend
Sub-issue of #28 to track progress for the Backend part:

- [X]  Create an API package with a common interface
- [ ]  Create a Driver package with a basic interface
- [x]  Create a “routes” package with common routes
- [ ]  Create a datasource package for handling connections

andresmgot:
Package for AWS SQL data sources that define common routes.

Epic at #28
Design [doc](https://docs.google.com/document/d/1V3SJGER5zIEQ_x-jHTE3gUIOvDMYqeYFsG4jxtTdux4/edit#)
Closes #31 

This PR implements, mainly, the `AWSDatasource` struct. The goal of this entity is to provide the methods to retrieve a database or an API instance, supporting multiple connections at the same time. The main concepts are:

 - Settings. This is the data source settings. It extends the basic AWS configuration (`backend.DataSourceInstanceSettings`) with settings specific to the data source.
 - API. An interface with the common methods to contact the data source API (implemented at #29).
 - Driver. Abstraction on top the the upstream sql.Driver. Defined to handle multiple connections. It's main purpose is to return databases.
 - Database. Upstream `sql.DB`.

The key point here is that the data source using this library should only worry about defining methods to initialize settings, APIs, databases... and forget about how to handle multiple connections.

### DIFF
--- a/pkg/sql/datasource/datasource.go
+++ b/pkg/sql/datasource/datasource.go
@@ -1,0 +1,168 @@
+package datasource
+
+import (
+	"database/sql"
+	"fmt"
+	"sync"
+
+	"github.com/grafana/grafana-aws-sdk/pkg/awsds"
+	"github.com/grafana/grafana-aws-sdk/pkg/sql/api"
+	"github.com/grafana/grafana-aws-sdk/pkg/sql/driver"
+	"github.com/grafana/grafana-aws-sdk/pkg/sql/models"
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/sqlds/v2"
+)
+
+// AWSDatasource stores a cache of:
+// - config: base configuration per datasource (only datasource ID)
+// - db: database connection per datasource and connection args
+// - driver: driver instance per datasource and connection args
+// - api: API instace per datasource and connection args
+// - sessionCache: cache used to create connections
+type AWSDatasource struct {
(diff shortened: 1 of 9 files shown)


pkg/sql/datasource/datasource.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/grafana/grafana-aws-sdk/pkg/awsds"
+	"github.com/grafana/grafana-aws-sdk/pkg/sql/api"
+	"github.com/grafana/grafana-aws-sdk/pkg/sql/driver"
+	"github.com/grafana/grafana-aws-sdk/pkg/sql/models"
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/sqlds/v2"
+)
+
+// AWSDatasource stores a cache of:
```
sarahzinger, line 16: love how well commented and well named this whole pr is! What would you think about a one sentence explanation for each of these terms, like "a driver is used xyz" just for folks new to these terms?

sarahzinger (approved):
This looks awesome to me!

pkg/sql/datasource/datasource.go
andresmgot: added more color to this comment.

PR #35, opened 2021-11-05
Remove frontend code

### LINKED ISSUE #26: Split the repo in two
It makes no sense to have both frontend and backend code in this repo since they don't share code, settings or version numbers. We should therefore move the frontend code and frontend tooling to a new repo - grafana-aws-sdk-js (or something). The backend code can be kept in this repo, but the repo should be renamed to grafana-aws-sdk-go.

Somewhat related to https://github.com/grafana/grafana-aws-sdk/issues/27.

### LINKED ISSUE #27: Setup CI/CD for this repo

andresmgot:
Frontend code is now at https://github.com/grafana/grafana-aws-sdk-react

Closes #26 
Closes #27 (CI/CD is already set in both repos)


andresmgot:
I think it's safe. In the unlikely event that we need an urgent update of the NPM package and the other repo is somehow broken we could always revert this commit but it doesn't hurt to be extra careful.

sarahzinger:
It's probably safe to remove but I think it'll be confusing if anyone new joins and is trying to understand where this code is coming from, +1 for keeping this pr open until we have an alternative ready to be consumed, but otherwise this is a ✅ from me!

ryantxu:
Any insights on the motivation for this?  is it just to have different repos for frontend+backend.  Is there anything wrong with having them together? 

obviously if we have two repos, the same code should not be duplicated in both places!

andresmgot:
> Any insights on the motivation for this? is it just to have different repos for frontend+backend. Is there anything wrong with having them together?

The major reason is that the two code bases are not related. Apart from that:
 - It's easier/faster to set CI (you only run go or typescript tests)
 - To make releases are consistent. If you go to the releases of this repo: https://github.com/grafana/grafana-aws-sdk/tags, those are go versions. In order to know what is the latest version of the NPM package, you need to go to the [package.json](https://github.com/grafana/grafana-aws-sdk/blob/main/package.json#L3) or to the [package view](https://github.com/grafana/grafana-aws-sdk/packages/639367) that by the way, they report different versions.

sunker:
Although not technically bound together, I guess they are related in the sense that they both interact with the data source json model. However, keeping them in one repo does not cater for safe, canonical interaction with the json model. 

@ryantxu are there any reasons for keeping them in the same repo? The only one I can think of is if we could use a shared schema for describing the data source jsondata model. That way we could theoretically build the backend and the frontend together. You know this better so correct me if I'm wrong, but I believe that's not something that the cue will be able to solve anytime soon? 

Here are my reasons for splitting the repo. None of them are absolutely necessary, but I think it's the combination of them that makes it the right decision: 
* It makes for cleaner code
* Consistent releases, having each repo produce one release/artifact only
* Should make CI slightly easier
* Makes local development slightly easier, not having to yarn link the go packages that you don't need in your frontend code
* Following the same structure as the google and the azure sdk, which should make life easier for our team since we're the ones maintaining all these sdk:s

ryantxu:
@sunker -- only reason is to avoid having lots of repos, and no real good reason to keep them separate.  github has gotten better at managing cross-repo and that cat is a bit out of the bag so :man_shrugging: 

I don't feel too strongly, just don't see any reason for it.  In parallel we trying to figure out how to have multiple projects/plugins in a single repo.

sunker (approved):
LGTM

PR #37, opened 2021-11-10
Chore: small improvements

andresmgot:
- SQL Datasource: Separate configuration storing from instance initialization.
- SQL Routes: Allow to send any resource, not only `[]string`

### DIFF
--- a/pkg/sql/datasource/datasource.go
+++ b/pkg/sql/datasource/datasource.go
@@ -29,10 +29,13 @@ type AWSDatasource struct {
 	db           sync.Map
 }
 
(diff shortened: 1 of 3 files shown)


iwysiu (approved):
LGTM! But what's the benefit from separating the initialization and the config? Do we ever want an unconfigured datasource?

andresmgot:
> LGTM! But what's the benefit from separating the initialization and the config? Do we ever want an unconfigured datasource?

Good question, this was a actually a bug. The instance should contain multiple configurations but the only way to add one was the initialization function (so it only held one).

PR #38, opened 2021-12-15
SQL data source: expose Init method

andresmgot:
Just some syntax sugar here. Rather than exposing the `StoreConfig` method, use an `Init` method instead.


sunker (approved):
LGTM

PR #39, opened 2021-12-15
Add method to attach an user agent to a session

andresmgot:
Context at https://github.com/grafana/athena-datasource/pull/97/files#r769580717


sunker (approved):
Nice!

PR #40, opened 2021-12-29
a11y: Add missing aria labels on connection config component

fridgepoet:
Accessibility improvement: tracked on https://github.com/grafana/grafana/issues/41206

Closes https://github.com/grafana/grafana/issues/42706

Hey @sarahzinger here's a go at adding aria-labels (indirectly?) for CloudWatch. Not sure at all if that's correct. Let me know what you think.

### DIFF
--- a/src/ConnectionConfig.tsx
+++ b/src/ConnectionConfig.tsx
@@ -68,6 +68,7 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
     <FieldSet label={skipHeader ? '' : 'Connection Details'} data-testid="connection-config">
       <InlineField
         label="Authentication Provider"
+        aria-label="Authentication Provider"
         labelWidth={28}
         tooltip="Specify which AWS credentials chain to use."
       >
@@ -180,6 +181,7 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
       )}
       <InlineField
         label="Default Region"
+        aria-label="Default Region"
         labelWidth={28}
         tooltip="Specify the region, such as for US West (Oregon) use ` us-west-2 ` as the region."
       >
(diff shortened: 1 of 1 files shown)


src/ConnectionConfig.tsx
```diff
@@ -72,6 +72,7 @@ export const ConnectionConfig: FC<ConnectionConfigProps> = (props: ConnectionCon
         tooltip="Specify which AWS credentials chain to use."
       >
         <Select
+          aria-label="Authentication Provider selection"
```
sarahzinger, line 75: looking good! I think 2 quick things:
- I don't think you need to add the word selection, I think it's sort of similar to this rule where you don't have to add the word `button`: https://www.aditus.io/aria/aria-label/#when-to-use
- I think it would be cool to explore if it's possible to use aria-labelledby instead since we already have a label next to these select dropdowns.

Let me know what you think! Thanks for looking at this 🎉

fridgepoet: > I think it would be cool to explore if it's possible to use aria-labelledby instead since we already have a label next to these select dropdowns.

I've tried to explore if this works, but I think it is only for any text and doesn't' pick up something from the `label` field.  Does that sound right?
However, now I'm puzzled that FastPass at one element reported no `placeholder` when there is...

yaelleC (commented):
lgtm 👍 

Just checking before approving: any reason why Fastpass is complaining about the select elems but not the inputs (Credential Profile Name for instance)? Or maybe were they not visible when you ran it?

fridgepoet:
Thanks @yaelleC you're right, the voiceover is not reading the titles of those other inputs. 
I think FastPass is passing them because they have placeholder attributes... 
What do you recommend? Should I just add more aria-labels?

yaelleC:
> Thanks @yaelleC you're right, the voiceover is not reading the titles of those other inputs. I think FastPass is passing them because they have placeholder attributes... What do you recommend? Should I just add more aria-labels?

Indeed! Turns out support of placeholders for screen readers is not consistent :( (see:[ https://www.davidmacd.com/blog/is-placeholder-accessible-label.html]( https://www.davidmacd.com/blog/is-placeholder-accessible-label.html)) So aria labels could be a good quick-fix solution and we can look into a better one is the issue you raise. 👍

sunker:
I believe this PR needs to be moved to the new frontend repo for aws - https://github.com/grafana/grafana-aws-sdk-react/.

fridgepoet:
Thanks
These aria-labels are already present in the new repo, closing!

PR #41, opened 2022-01-11
Remove SQL db cache

andresmgot:
The interface method `GetDB` was returning a cached version of the db connection if the same datasource ID and connections args were passed. This is wrong because the db connection also depend on the datasource settings. This means that even after modifying the datasource settings, calling `GetDB` would return a db connection with wrong parameters.

To fix that, we would need to also take into consideration the datasource settings to store the db connection (or expose a different method that always creates a db connection) but truth is that we don't need to cache this connection at all. The `GetDB` method is only called in the [`Connect` method of the datasource](https://github.com/grafana/athena-datasource/blob/main/pkg/athena/datasource.go#L54) and this `Connect` method is only called when there is an actual need of a new connection:
 - When saving a datasource.
 - When running a query against a datasource for the first time.
 - When running a query with different connection arguments for the first time.

Because of that, I am removing that connection from the cache. Also note that the [underlying AWS session](https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/sessions.go#L107) (which handle the underlying connections) is already caching sessions so even if we call `GetDB` several times with the same arguments (e.g. clicking "Save datasource" several times), only one connection will be set up.

Fixes https://github.com/grafana/athena-datasource/issues/105 (once released)

### DIFF
--- a/pkg/sql/datasource/datasource.go
+++ b/pkg/sql/datasource/datasource.go
@@ -26,7 +26,6 @@ type AWSDatasource struct {
 	config       sync.Map
 	api          sync.Map
 	driver       sync.Map
-	db           sync.Map
 }
 
 func New() *AWSDatasource {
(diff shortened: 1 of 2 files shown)


sarahzinger (approved):
makes sense to me!

PR #42, opened 2022-01-13
Use data source instance HTTP options in AWS sessions

andresmgot:
`GetSession` now accepts the data source configuration and it's used to obtain HTTP client options.

We were already using different parameters: the region, the default region, the AWS settings, the user agent... I gathered all those in a single struct that is now the only parameter for `GetSession`. This simplification implies a breaking change though.

First step of [#39089](https://github.com/grafana/grafana/issues/39089)

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/httpclient"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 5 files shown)


pkg/awsds/sessions.go
```diff
@@ -234,17 +258,3 @@ func (sc *SessionCache) GetSession(region string, s AWSDatasourceSettings) (*ses
...
-
-func GetSessionWithDefaultRegion(sessionCache *SessionCache, settings AWSDatasourceSettings) (*session.Session, error) {
-	region := settings.DefaultRegion
-	if settings.Region != "" {
-		region = settings.Region
-	}
-	return sessionCache.GetSession(region, settings)
-}
-
-func WithUserAgent(sess *session.Session, name string) {
```
iwysiu, line 246: Good call adding this to the constructor!

iwysiu (approved):
lgtm!

PR #43, opened 2022-01-17
Session config: use HTTP client provided

andresmgot:
Rather than using the HTTP config and create a client, use the given client.

(This feature is not in use yet).

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -2,6 +2,7 @@ package awsds
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"strconv"
 	"strings"
@@ -9,7 +10,6 @@ import (
 	"time"
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
-	"github.com/grafana/grafana-plugin-sdk-go/backend/httpclient"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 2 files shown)


marefr (approved):
LGTM

Reg. `UserAgentName` how are you using that? If you start using `httpClient` from `httpClientProvider` the user agent might be set to `Grafana vX.Y.Z` due to this [middleware](https://github.com/grafana/grafana/blob/c6183394820cf02dc0410a465bcceaf0be24a0fb/pkg/infra/httpclient/httpclientprovider/http_client_provider.go#L25). Might be good to double-check that works as expected when changing the default AWS SDK http client

andresmgot:
> Reg. `UserAgentName` how are you using that? If you start using `httpClient` from `httpClientProvider` the user agent might be set to `Grafana vX.Y.Z` due to this [middleware](https://github.com/grafana/grafana/blob/c6183394820cf02dc0410a465bcceaf0be24a0fb/pkg/infra/httpclient/httpclientprovider/http_client_provider.go#L25). Might be good to double-check that works as expected when changing the default AWS SDK http client

Good point. It's being used here: 

https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/sessions.go#L246

I checked with the debugger, that handler function is executed before the middleware. Since the middleware is [checking if the agent is already set](https://github.com/grafana/grafana/blob/c6183394820cf02dc0410a465bcceaf0be24a0fb/pkg/infra/httpclient/httpclientprovider/user_agent_middleware.go#L21), it skips that step.

PR #45, opened 2022-01-27
Fix ec2 with endpoint connection

iwysiu:
Fixes grafana/support-escalations#1531

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -168,10 +168,6 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 		cfgs = append(cfgs, regionCfg)
 	}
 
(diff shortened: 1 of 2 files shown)


andresmgot (approved):
LGTM!

sunker (approved):
Yes looks good! Would be great if we could test this in ec2 (a workspace in aws would be even better) using cloudwatch before releasing a new version of the sdk. 

Steps to reproduce in the escalation, and you should be able to use endpoint `monitoring.us-east-2.amazonaws.com` while testing.

sunker:
Nice @iwysiu! We also need to bump the sdk version in all plugins that use it.

andresmgot:
we may want to work on https://github.com/grafana/grafana-aws-sdk/issues/46 before updating all the plugins

sunker:
Yeah makes sense

PR #48, opened 2022-02-03
SigV4: Guarantee consistency of signed headers

wbrowne:
The main issue this PR addresses is previously we allowed a fixed set of headers to be signed. This leads to problems with `Content-Length` in particular as `Content-Length: 0` set as part of a GET request (valid) can cause signature issues depending on what infrastructure the request passes through. We can't choose to omit `Content-Length` as this will be problematic for `POST` requests. So now instead, parameters from the request are copied to a new request, which is then signed in its most minimal form. Subsequently from there we copy over headers from the original request to the new signed request. This approach is modelled off of AWS' approach - https://github.com/awslabs/aws-sigv4-proxy

Changes:

- Reconstructs signer only when necessary
- Only signs minimal amount of headers (`Host` and `X-Amz-Date`) - using the approach followed [here](https://github.com/awslabs/aws-sigv4-proxy) as a reference
- Fixes badly formatted log lines
- Added support for debugging bad signature scenarios by requiring a `Logger` implementation

Example of debugging output from Grafana logs:

```
INFO[02-03|12:00:13] Request dump                             logger=sigv4 dump="GET /api/datasources/proxy/41/.kibana_1/_mapping HTTP/1.1\r\nHost: <INSTANCE>.us-east-2.es.amazonaws.com\r\nAccept: application/json, text/plain, */*\r\nAccept-Encoding: gzip, deflate, br\r\nAccept-Language: en-GB,en-US;q=0.9,en;q=0.8\r\nSec-Ch-Ua: \" Not;A Brand\";v=\"99\", \"Google Chrome\";v=\"97\", \"Chromium\";v=\"97\"\r\nSec-Ch-Ua-Mobile: ?0\r\nSec-Ch-Ua-Platform: \"macOS\"\r\nSec-Fetch-Dest: empty\r\nSec-Fetch-Mode: cors\r\nSec-Fetch-Site: same-origin\r\nUser-Agent: Grafana/8.4.0-pre\r\nX-Forwarded-For: ::1, ::1\r\nX-Grafana-Nocache: true\r\nX-Grafana-Org-Id: 1\r\n\r\n" stage=pre-signature
[INFO] [02-03|12:00:13]                                       logger=sigv4 DEBUG: Request Signature:
---[ CANONICAL STRING  ]-----------------------------
GET
/.kibana_1/_mapping

host:<INSTANCE>.us-east-2.es.amazonaws.com
x-amz-date:20220203T120013Z

host;x-amz-date
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
---[ STRING TO SIGN ]--------------------------------
AWS4-HMAC-SHA256
20220203T120013Z
20220203/us-east-2/es/aws4_request
afa91bf5f8ef0049023b8e7481db2bb07548c4f12e867ee63517e72cad3cd42a
-----------------------------------------------------=(MISSING)
INFO[02-03|12:00:13] Request dump                             logger=sigv4 dump="GET /.kibana_1/_mapping HTTP/1.1\r\nHost: <INSTANCE>.us-east-2.es.amazonaws.com\r\nAccept: application/json, text/plain, */*\r\nAccept-Encoding: gzip, deflate, br\r\nAccept-Language: en-GB,en-US;q=0.9,en;q=0.8\r\nAuthorization: AWS4-HMAC-SHA256 Credential=<ACCESS_KEY_ID>/20220203/us-east-2/es/aws4_request, SignedHeaders=host;x-amz-date, Signature=25f1b87b10d7eabadc18fcfdd7532b7a76578a663218e0d8275b28abf9490212\r\nSec-Ch-Ua: \" Not;A Brand\";v=\"99\", \"Google Chrome\";v=\"97\", \"Chromium\";v=\"97\"\r\nSec-Ch-Ua-Mobile: ?0\r\nSec-Ch-Ua-Platform: \"macOS\"\r\nSec-Fetch-Dest: empty\r\nSec-Fetch-Mode: cors\r\nSec-Fetch-Site: same-origin\r\nUser-Agent: Grafana/8.4.0-pre\r\nX-Amz-Date: 20220203T120013Z\r\nX-Forwarded-For: ::1, ::1\r\nX-Grafana-Nocache: true\r\nX-Grafana-Org-Id: 1\r\n\r\n" stage=post-signature
```

Associated issue: https://github.com/grafana/support-escalations/issues/1028


andresmgot (commented):
Mostly looks good to me but I feel I don't have full context so letting @marefr to do a final review. BTW, should this be associated to some issue?

pkg/sigv4/sigv4.go
```diff
@@ -70,57 +71,107 @@ func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	VerboseMode() bool
+}
+
 // New instantiates a new signing middleware with an optional succeeding
 // middleware. The http.DefaultTransport will be used if nil
-func New(config *Config, next http.RoundTripper) http.RoundTripper {
-	// Need to delay fetching auth settings until env vars have had a chance to propagate
-	if authSettings == nil {
-		authSettings = awsds.ReadAuthSettingsFromEnvironmentVariables()
+func New(cfg *Config, next http.RoundTripper, logger Logger) (http.RoundTripper, error) {
```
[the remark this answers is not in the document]
wbrowne: Yeah it's not great - I just amended to introduce a second signature so as not to have a breaking change. Alternatively it might look better with a variadic `opts` argument of some sort perhaps. 

I'm not too keen on adjusting the Config struct as that's dedicated for AWS specific config for authentication, so doesn't feel at home there IMO.

pkg/sigv4/sigv4.go
```diff
@@ -60,6 +54,13 @@ type Config struct {
 	Region        string
 }
 
+func (c Config) asSha256() string {
+	h := sha256.New()
+	h.Write([]byte(fmt.Sprintf("%v", c)))
```
marefr, line 59: Lint error, need to handle error

pkg/awsds/settings.go
```diff
@@ -34,20 +34,20 @@ func (at AuthType) String() string {
 	}
 }
 
-func ToAuthType(authType string) AuthType {
+func ToAuthType(authType string) (AuthType, error) {
```
marefr, line 37: Is this a breaking change? And/or only used internally?

pkg/sigv4/sigv4.go
```diff
@@ -70,57 +71,112 @@ func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return rt(r)
 }
 
+type Logger interface {
```
marefr, line 74: Want to use the logger interface from SDK, https://github.com/grafana/grafana-plugin-sdk-go/blob/a62a98d0ea262e968f84970dc9fce2ae4969da34/backend/log/log.go#L20-L26, instead? 

The request specific logger can be implemented on top I suppose.

Update: Ohh I see you provide this to aws SDK now so understand it requires a certain interface. The verbose part as well, probably needs some special handling as well so that we don't expose sensible information in logs.

pkg/sigv4/sigv4.go
```diff
@@ -70,57 +71,112 @@ func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
+func New(cfg *Config, next http.RoundTripper) (http.RoundTripper, error) {
+	return newRoundTripper(cfg, next, &nopLogger{})
+}
 
+func NewWithLogger(cfg *Config, next http.RoundTripper, logger Logger) (http.RoundTripper, error) {
+	return newRoundTripper(cfg, next, logger)
+}
+
+func newRoundTripper(cfg *Config, next http.RoundTripper, logger Logger) (http.RoundTripper, error) {
+	if err := validateConfig(cfg); err != nil {
```
marefr, line 91: Maybe should check that logger is not nil?

pkg/sigv4/sigv4.go
```diff
@@ -16,32 +18,24 @@ import (
...
-	"securitytenant":  {},
-	"sgtenant":        {},
-	"kbn-xsrf":        {},
-}
-
 var (
-	authSettings *awsds.AuthSettings = nil
+	signerCache = struct {
+		cache map[string]*v4.Signer
+		sync.RWMutex
```
marefr, line 28: Might consider using https://pkg.go.dev/sync#Map

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

marefr (changes requested):
In general looks good, but added some comments.

pkg/awsds/settings.go
wbrowne: Yeah - currently only used internally. Was specifically introduced as a helper for SigV4 so it looks like it's the only thing using it AFAICT

pkg/sigv4/sigv4.go
```diff
@@ -70,57 +69,112 @@ func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
-	// Need to delay fetching auth settings until env vars have had a chance to propagate
-	if authSettings == nil {
-		authSettings = awsds.ReadAuthSettingsFromEnvironmentVariables()
+func New(cfg *Config, next http.RoundTripper) (http.RoundTripper, error) {
+	return newRoundTripper(cfg, next, &nopLogger{})
+}
+
+func NewWithLogger(cfg *Config, next http.RoundTripper, logger Logger) (http.RoundTripper, error) {
+	return newRoundTripper(cfg, next, logger)
+}
```
andresmgot, line 86: will the logger really change in different calls? it seems to me that this logger parameter is only defined to use a different logger in the tests. If that's the case, I would define a default logger as a package variable so the caller does not need to use this new function and doesn't have to bother to learn about how this logger works internally.

wbrowne: The main idea with the logger is to provide us an easier way to debug SigV4 issues as we don't have anything in place right now. With that, we need some form of  control that so that we can only enable that when appropriate - as for example we don't want to always do HTTP request dumps. Given that it's an interface, the caller can decide what to do with it, based on however logging is done there.

The existing logger from the plugin SDK doesn't provide anything useful unfortunately. Furthermore, this new interface also fits the shape of the interface expected by the AWS logger IE
```golang
var signerOpts = func(s *v4.Signer) {
	if logger.VerboseMode() {
		s.Logger = logger
		s.Debug = aws.LogDebugWithSigning
	}
}
```

Thinking about it differently, maybe instead we could have:

```golang
func New(cfg *Config, next http.RoundTripper, opts ...SigV4Options) (http.RoundTripper, error) {
	return newRoundTripper(cfg, next, opts.debugEnabled)
}

```

`debugEnabled` can be our control to enable logging within the AWS SDK (using aws.LogDebugWithSigning) and to also decide whether or not to dump the requests. And we can use plugin SDK logger for our logging needs.

🤔  WDYT?

andresmgot: got it. So my question now is, would the logger level differ between the default logger and this logger? (for example setting the default log level to "info" and this one to "debug"). If the answer is "no, both will be the same", then, is there a better way to get that log level without changing the interface of this method? (something like `backend.Logger.Level()` or alike). If the answer is no again then I agree with your proposed interface change. Thanks for checking this!

wbrowne: Technically they could be different, but I think they should not differ. 

Ah! Great tip - that just reminded me that this [exact thing was added to the plugin SDK](https://github.com/grafana/grafana-plugin-sdk-go/pull/435). Let me give that a shot

pkg/awsds/settings.go
marefr: Okay :+1:

pkg/sigv4/sigv4.go
marefr: But level is always debug by default for backend plugins so I think a specific verbose config for sigv4 is still needed. Talked about it in my update here https://github.com/grafana/grafana-aws-sdk/pull/48#discussion_r798752336

pkg/sigv4/sigv4.go
```diff
@@ -60,6 +53,16 @@ type Config struct {
 	Region        string
 }
 
+func (c Config) asSha256() string {
+	h := sha256.New()
+	_, err := h.Write([]byte(fmt.Sprintf("%v", c)))
+	if err != nil {
+		return ""
```
marefr, line 60: Nit. Not sure in what way this could fail, but feels like the error should be returned rather than swallowed.

pkg/sigv4/sigv4.go
```diff
@@ -133,94 +172,119 @@ func (m *middleware) signer() (*v4.Signer, error) {
 		return nil, fmt.Errorf("attempting to use an auth type for SigV4 that is not allowed: %q", authType.String())
 	}
 
-	if m.config.AssumeRoleARN != "" && !authSettings.AssumeRoleEnabled {
+	if cfg.AssumeRoleARN != "" && !authSettings.AssumeRoleEnabled {
 		return nil, fmt.Errorf("attempting to use assume role (ARN) for SigV4 which is not enabled")
 	}
 
+	var signerOpts = func(s *v4.Signer) {
+		if plog.Level() == log.Debug {
```
marefr, line 180: Already left a comment regarding this. Default level is debug for backend plugins so this is probably not what we want given the sensitive information that might be logged. Verbose config was more suitable I think.

marefr (commented):
Great, but some minor things left.

pkg/sigv4/sigv4.go
wbrowne: Maybe the `trace` level in that case?

marefr: Yeah might work

marefr: Only thing to keep in mind is that Grafana will log traces with debug level

wbrowne: Yeah.. it's a weird side effect. Okay let me look at the passing opts instead 🕺

pkg/sigv4/sigv4.go
```diff
@@ -133,94 +182,119 @@ func (m *middleware) signer() (*v4.Signer, error) {
...
 	}
+	plog.Debug("Request dump", append([]interface{}{"dump", string(dump)}, args...)...)
+}
+
+type awsLoggerAdapter struct {
+	logger log.Logger
+}
+
+func (a awsLoggerAdapter) Log(args ...interface{}) {
+	a.logger.Info("AWS debug log", args...)
```
andresmgot, line 299: this is new, right? do you want that to appear in all the logs?

andresmgot (approved):
thanks for checking the changes!

marefr (approved):
LGTM, but needs to fix failing test

PR #49, opened 2022-02-17
Use non opt-in region when obtaining tokens for assumed roles

### LINKED ISSUE #46: Using Assume Role ARN with opt-in regions gives 403
Some AWS regions are opt-in regions, which means they can't be accessed until they are explicitly turned on (see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions)

When using the Assume Role ARN field, credentials from one account (Account A) can be used to retrieve data from another account (Account B). Currently there exists a bug when trying to load data from an opt-in region in Account B will give 
```
"error":"failed to query data: InvalidClientTokenId: The security token included in the request is invalid.\n\tstatus code: 403
```

This is because the region selected in the configuration is used both for the call to STS to assume the role and the call to the data source (e.g. cloudwatch). This can be fixed by replacing the region used here https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/sessions.go#L167 with a non-opt-in region (e.g. us-east-1) when the configured region is an opt-in region and Assume Role ARN is configured. Then the regionCfg used here https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/sessions.go#L227 should use the configured region.

Note the environment variable `AWS_STS_REGIONAL_ENDPOINTS=regional` must be set to get credentials that can be used in an opt-in region.

Steps to reproduce:
1. In one AWS account (Account A) Create a user with AssumeRole permissions
e.g.
```
{
    "Version": "2012-10-17",
    "Statement": {
        "Effect": "Allow",
        "Action": [
            "sts:AssumeRole"
        ],
        "Resource": "*"
    }
}
```
2. In another AWS account (Account B) opt in to an opt-in region (e.g. me-south-1)
3. In Account B create a role, attach a CloudWatchReadOnlyAccess permission policy and a trust policy for Account A
e.g.
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::<AccountA>:root"
      },
      "Action": "sts:AssumeRole",
      "Condition": {}
    }
  ]
}
```
4. Create a CloudWatch data source
5. Provide an Access Key and Secret Key for the user in step 1
6. Provide the role ARN from step 3 in Assume Role ARN
7. Select `me-south-1` as Default Region
8. Save & Test

andresmgot:
Fixes #46 

When assuming a role in another account, the region targeted may not exist in the current account ([opt-in regions](https://docs.aws.amazon.com/general/latest/gr/rande-manage.html)) so we need to use one that we know it will exists.

cc/ @robbierolin

### DIFF
--- a/pkg/awsds/sessions_test.go
+++ b/pkg/awsds/sessions_test.go
@@ -1,7 +1,6 @@
 package awsds
 
 import (
-	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -142,20 +141,13 @@ func TestNewSession_AssumeRole(t *testing.T) {
 
 	t.Run("Assume role is enabled with an opt-in region", func(t *testing.T) {
 		resetEnvironmentVariables()
-		prevF := newSTSCredentials
+		fakeNewSTSCredentials := newSTSCredentials
 		newSTSCredentials = func(c client.ConfigProvider, roleARN string,
 			options ...func(*stscreds.AssumeRoleProvider)) *credentials.Credentials {
 			sess := c.(*session.Session)
-			if *sess.Config.Region != "us-east-1" {
-				// Verify that we are using the well-known region
-				t.Errorf(fmt.Sprintf("unexpected region %s", *sess.Config.Region))
-			}
-			p := &stscreds.AssumeRoleProvider{RoleARN: roleARN}
-			for _, o := range options {
-				o(p)
-			}
-
-			return credentials.NewCredentials(p)
+			// Verify that we are using the well-known region
+			assert.Equal(t, *sess.Config.Region, "us-east-1")
+			return fakeNewSTSCredentials(c, roleARN, options...)
 		}
 		settings := AWSDatasourceSettings{
 			AssumeRoleARN: "test",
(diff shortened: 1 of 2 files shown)


pkg/awsds/sessions_test.go
```diff
@@ -138,6 +139,38 @@ func TestNewSession_AssumeRole(t *testing.T) {
...
 		assert.Empty(t, diff)
 	})
+
+	t.Run("Assume role is enabled with an opt-in region", func(t *testing.T) {
+		resetEnvironmentVariables()
+		prevF := newSTSCredentials
+		newSTSCredentials = func(c client.ConfigProvider, roleARN string,
+			options ...func(*stscreds.AssumeRoleProvider)) *credentials.Credentials {
+			sess := c.(*session.Session)
+			if *sess.Config.Region != "us-east-1" {
```
fridgepoet, line 149: Looks like we can `assert` the region is `us-east-1` here

pkg/awsds/sessions_test.go
```diff
@@ -138,6 +139,38 @@ func TestNewSession_AssumeRole(t *testing.T) {
...
+				o(p)
+			}
+
+			return credentials.NewCredentials(p)
+		}
+		settings := AWSDatasourceSettings{
+			AssumeRoleARN: "test",
+			Region:        "me-south-1",
+		}
+		os.Setenv(AllowedAuthProvidersEnvVarKeyName, "default")
```
fridgepoet, line 164: I guess these env variables are needed for the session? do we need to unset them after the test?

fridgepoet: Nevermind I see the first line of the test

PR #52, opened 2022-04-20
Fix session creation when assuming a role in us-gov regions

andresmgot:
Issue: https://github.com/grafana/athena-datasource/issues/145

We have detected a regression caused by #49. When targeting a us-gov region, the AWS session should be created for that region, in other case, the token obtained will be invalid. More details about differences with Gov regions here: https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-differences.html

This PR hardcodes the list of opt-in regions that are available today. This has the disadvantage that at some point this list will become outdated. I tried other other possibilities:

 - Enumerating the regions using the [endpoints package](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/). The problem with this solution is that the list returned doesn't contain information about if a region is optional or not.
 - Using the [`DescribeRegions` of the ec2 client](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeRegions). The problem with this is that it is hitting the AWS API, which requires a session but we need this check _before_ the session is created.

### DIFF
--- a/pkg/awsds/sessions_test.go
+++ b/pkg/awsds/sessions_test.go
@@ -146,7 +146,7 @@ func TestNewSession_AssumeRole(t *testing.T) {
 			options ...func(*stscreds.AssumeRoleProvider)) *credentials.Credentials {
 			sess := c.(*session.Session)
 			// Verify that we are using the well-known region
-			assert.Equal(t, *sess.Config.Region, "us-east-1")
+			assert.Equal(t, "us-east-1", *sess.Config.Region)
 			return fakeNewSTSCredentials(c, roleARN, options...)
 		}
 		settings := AWSDatasourceSettings{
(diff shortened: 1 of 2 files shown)


aangelisc (approved):
LGTM - a happy medium I think 😊 

We could alternatively maintain a list of the default regions but I think the overhead here is negligible to update when/if a new opt-in region is released.

pkg/awsds/sessions_test.go
```diff
@@ -163,6 +163,31 @@ func TestNewSession_AssumeRole(t *testing.T) {
...
 	})
+
+	t.Run("Assume role is enabled with a gov region", func(t *testing.T) {
+		resetEnvironmentVariables()
+		fakeNewSTSCredentials := newSTSCredentials
+		newSTSCredentials = func(c client.ConfigProvider, roleARN string,
+			options ...func(*stscreds.AssumeRoleProvider)) *credentials.Credentials {
+			sess := c.(*session.Session)
+			// Verify that we are using the well-known region
+			assert.Equal(t, *sess.Config.Region, "us-gov-east-1")
```
fridgepoet, line 174:
```suggestion
			assert.Equal(t, "us-gov-east-1", *sess.Config.Region)
```
should be in this order for this library. (seems to be mistakenly in reverse order in other tests too)

PR #55, opened 2022-06-06
Add function to marshal errors correctly

aangelisc:
Forms a part of grafana/grafana#50238. 

Fixes grafana/athena-datasource#141 as an error will be correctly parsed and returned as seen in the screenshot below:
image

### DIFF
--- a/pkg/sql/routes/routes.go
+++ b/pkg/sql/routes/routes.go
@@ -35,17 +35,28 @@ func ParseBody(body io.ReadCloser) (sqlds.Options, error) {
 	return reqBody, nil
 }
 
+func marshalError(rw http.ResponseWriter, err error, code int) {
+	rw.WriteHeader(code)
+	errBytes, err := json.Marshal(err.Error())
+	if err != nil {
+		log.DefaultLogger.Error(err.Error())
+		rw.WriteHeader(http.StatusInternalServerError)
+		jsonErr, _ := json.Marshal(err)
+		Write(rw, jsonErr)
+		return
+	}
+	Write(rw, errBytes)
+}
+
 func SendResources(rw http.ResponseWriter, res interface{}, err error) {
 	if err != nil {
 		rw.WriteHeader(http.StatusBadRequest)
-		Write(rw, []byte(err.Error()))
+		marshalError(rw, err, http.StatusBadRequest)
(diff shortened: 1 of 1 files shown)


pkg/sql/routes/routes.go
```diff
@@ -35,17 +35,28 @@ func ParseBody(body io.ReadCloser) (sqlds.Options, error) {
...
+		jsonErr, _ := json.Marshal(err)
+		Write(rw, jsonErr)
+		return
+	}
+	Write(rw, errBytes)
+}
+
 func SendResources(rw http.ResponseWriter, res interface{}, err error) {
 	if err != nil {
 		rw.WriteHeader(http.StatusBadRequest)
```
andresmgot, line 53: you don't need this line now

pkg/sql/routes/routes.go
```diff
@@ -35,17 +35,28 @@ func ParseBody(body io.ReadCloser) (sqlds.Options, error) {
 	return reqBody, nil
 }
 
+func marshalError(rw http.ResponseWriter, err error, code int) {
+	rw.WriteHeader(code)
+	errBytes, err := json.Marshal(err.Error())
+	if err != nil {
+		log.DefaultLogger.Error(err.Error())
```
andresmgot, line 42: since this is executed for all the errors, it's better to use the `.Debug` function

pkg/sql/routes/routes.go
```diff
@@ -35,17 +35,28 @@ func ParseBody(body io.ReadCloser) (sqlds.Options, error) {
...
+		rw.WriteHeader(http.StatusInternalServerError)
+		jsonErr, _ := json.Marshal(err)
```
andresmgot, line 44: you need to handle the error here (and probably just log it). If there were an error, `jsonErr` will be nil and it would probably panic so you would need to `return` early

andresmgot (commented):
Good find! I have some small suggestions.

pkg/sql/routes/routes.go
aangelisc: I believe I do need this line, I've renamed the variables for clarity

andresmgot (approved):
LGTM, thanks!

PR #56, opened 2022-06-09
Include external id in session cache key

mtanda:
If external id is specified, API call from datasource with invalid external id, should fail.
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

Currently, Grafana doesn't include external id in cache key.
It may cause issue.

If Grafana has two datasource settings.
- datasource A with valid external ID
- datasource B with invalid external ID

Grafana will create one session cache from these two datasource setting.
If Grafana cache datasource A, API call which using datasource B will success (it should fail).

I suggest to include external id in cache key.

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -146,7 +146,7 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 
 	bldr := strings.Builder{}
 	for i, s := range []string{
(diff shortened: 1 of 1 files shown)


andresmgot (commented):
Thanks for the PR @mtanda, we are aware of this issue. It happens with more fields, not only the ExternalID (also the SecretKey, for example). We have discussed this in https://github.com/grafana/athena-datasource/issues/144#issuecomment-1140603589 but we didn't get to it yet. Our idea is to use a hash of all the settings so we catch any change.

PR #57, opened 2022-06-22
Chore: Allow logger to be overridden

marefr:
By assigning `backend.Logger` to a variable it makes it impossible to override the logger implementation when grafana-aws-sdk is used in Grafana itself for the core CloudWatch datasource. This removes the assignment of `backend.Logger` to a variable.

### DIFF
--- a/pkg/sigv4/sigv4_test.go
+++ b/pkg/sigv4/sigv4_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
 	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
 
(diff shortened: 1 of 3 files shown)


pkg/sigv4/sigv4_test.go
```diff
@@ -141,13 +142,13 @@ func TestNew(t *testing.T) {
 		cfg := &Config{AuthType: "ec2_iam_role"}
 
 		// Mock logger
-		origLogger := plog
+		origLogger := backend.Logger
 		t.Cleanup(func() {
-			plog = origLogger
+			backend.Logger = origLogger
```
andresmgot, line 147: I am surprised you can mock an exported variable like this but the tests are passing so it should be fine :+1:

andresmgot (approved):
LGTM

wbrowne (approved):
😁

PR #61, opened 2022-08-30
Long Running Queries

kevinwcyu:
Part of #64


go.mod
```diff
@@ -64,5 +64,7 @@ require (
 	google.golang.org/grpc v1.41.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/grafana/sqlds/v2 => ../sqlds
```
kevinwcyu, line 70: This is just a temporary change to be able to test locally.

sunker (approved):
Great stuff! Just added a few questions. 

It's a shame external plugins can't expose prometheus metrics because it would have been interesting to count the frequence of a status so we could monitor the query cycle.

pkg/awsds/asyncDatasource.go
```diff
@@ -0,0 +1,185 @@
...
+		if len(res) == 0 {
+			res = append(res, &data.Frame{})
+		}
+		res[0].Meta.Custom = customMeta
+	}
+	return res, err
+}
+
+func queryAsync(ctx context.Context, conn *sql.DB, converters []sqlutil.Converter, fillMode *data.FillMissing, q *AsyncQuery) (data.Frames, error) {
+	return sqlds.QueryDB(ctx, conn, converters, fillMode, &q.Query, sql.NamedArg{Name: "queryID", Value: q.QueryID})
```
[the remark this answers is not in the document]
iwysiu: We could to some extent, but we wouldn't be saving much code. The part of `handleQuery` we could reuse is handling the connection and the first `QueryDB` call. We need to do the interpolation here to make sure we look up the right query and `handleQuery`'s second `QueryDB` call uses different connection logic.

pkg/awsds/asyncDatasource.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Apply the default FillMode, overwritting it if the query specifies it
+	driverSettings := ds.SQLDatasource.DriverSettings()
+	fillMode := driverSettings.FillMode
+	if q.FillMissing != nil {
+		fillMode = q.FillMissing
+	}
+
+	if driverSettings.Timeout != 0 {
+		tctx, cancel := context.WithTimeout(ctx, driverSettings.Timeout)
```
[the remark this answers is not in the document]
sunker: Yeah, let's remove it from the async flow then. It will only add confusion.

pkg/awsds/asyncDatasource.go
```diff
@@ -0,0 +1,185 @@
...
+	if !status.Finished() {
+		return data.Frames{
+			{Meta: &data.FrameMeta{
+				ExecutedQueryString: q.RawSQL,
+				Custom:              customMeta},
+			},
+		}, nil
+	}
+
+	conn, err := ds.driver.Connect(ds.connSettings, q.ConnectionArgs)
```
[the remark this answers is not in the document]
sunker: You may want to test this manually and make sure this works as expected. You can for example add two query rows in a panel and make sure they use different regions. Then each of the query rows should have one connection each.

pkg/awsds/asyncDatasource.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/instancemgmt"
+	"github.com/grafana/grafana-plugin-sdk-go/data"
+	"github.com/grafana/grafana-plugin-sdk-go/data/sqlutil"
+	"github.com/grafana/sqlds/v2"
+)
+
+type AsyncAWSDatasource struct {
+	*sqlds.SQLDatasource
+	asyncDB              AsyncDB
```
andresmgot, line 19: this is why I think this is not an "async datasource" but an "async sql data source" since we are still tied to the database architecture of sqlds

pkg/awsds/asyncDatasource.go
```diff
@@ -0,0 +1,164 @@
...
+	return &AsyncAWSDatasource{
+		SQLDatasource:        sqlDs,
+		driver:               driver,
+		sqldsQueryDataHander: sqlDs.QueryData,
+	}
+}
+
+func getQueryFlow(query backend.DataQuery) string {
+	q, _ := GetQuery(query)
+	if q.Meta.QueryFlow == "async" {
```
andresmgot, line 36: this is because its the only way of using a "feature flag" for plugins, right? If that's the case, we can add a comment here explaining the situation.

pkg/awsds/asyncDatasource.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (ds *AsyncAWSDatasource) NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.Instance, error) {
+	var err error
+	ds.connSettings = settings
+	ds.asyncDB, err = ds.driver.GetAsyncDB(settings, nil)
+	if err != nil {
+		return nil, err
+	}
+	_, err = ds.SQLDatasource.NewDatasource(settings)
```
andresmgot, line 49: this is a bit weird, you are creating a data source using sqlds but not using it?

pkg/awsds/asyncDatasource.go
andresmgot: also a nit but you can just define this as `isAsyncFlow(...) bool` since there are only two values returned.

pkg/awsds/utils.go
```diff
@@ -37,3 +39,23 @@ func GetUserAgentString(name string) string {
...
 }
+
+func getDatasourceUID(settings backend.DataSourceInstanceSettings) string {
+	datasourceUID := settings.UID
+	// Grafana < 8.0 won't include the UID yet
+	if datasourceUID == "" {
+		datasourceUID = fmt.Sprintf("%d", settings.ID)
+	}
+	return datasourceUID
+}
```
andresmgot, line 50: this function is already defined in sqlds, you may want to expose that (or simply assume that settings.UID is defined, I don't think we support Grafana < 8.0 anyway.

pkg/sql/datasource/datasource.go
```diff
@@ -130,6 +150,35 @@ func (s *AWSDatasource) GetDB(
...
+// GetAsyncDB returns a sqlds.AsyncDB. It will use the loader functions to initialize the required
+// settings, API and driver and finally create a DB.
+func (s *AWSDatasource) GetAsyncDB(
+	id int64,
+	options sqlds.Options,
+	settingsLoader models.Loader,
+	apiLoader api.Loader,
+	driverLoader asyncDriver.Loader,
+) (awsds.AsyncDB, error) {
+	// TODO: Reuse stuff from GetDB
```
andresmgot, line 162: is this TODO still relevant?

andresmgot (approved):
Good job! Just a few minor comments. Feel free to ignore

pkg/awsds/asyncDatasource.go
andresmgot: okay, I think we got the naming there wrong. It should be `func (ds *SQLDatasource) Init(settings backend.DataSourceInstanceSettings) (error)`, also because we already have a `NewDatasource` that is really generating a new data source.

iwysiu: Yeah, it's badly named 😞  Unfortunately I don't think the `sqlds` function can be renamed at this point without a breaking change.

andresmgot: well, you just released that so nobody is using it but here so you have some flexibility if you want to modify it. In any case releases are free and numbers are infinite so there is no problem doing breaking changes :) (at least for things like this)

iwysiu: Unfortunately it was already being exported `NewDatasource` before the long running queries changes 😞 I could deprecate it, but again, if we're planning on just not using it in the future there doesn't seem to be that much of a point.

PR #69, opened 2023-01-25
Update CODEOWNERS

fridgepoet:
Adding AWS Plugins as codeowners except for sig4


vickyyyyyyy (approved):
Thanks for the catch! 🎉

PR #71, opened 2023-02-09
Pass query args to GetAsyncDB

kevinwcyu:
Work for Athena query result reuse work.

This passes the ConnectionArgs to the Datasource's `GetAsyncDB` method. The `ConnectionArgs` are specific to a query. Currently, synchronous queries already get the connection args (this is handled in sqlds https://github.com/grafana/sqlds/blob/f35c5a62f66c02aef63b260c1e0a25568ae698b6/datasource.go#L169).

### DIFF
--- a/pkg/awsds/asyncDatasource.go
+++ b/pkg/awsds/asyncDatasource.go
@@ -115,8 +115,9 @@ func (ds *AsyncAWSDatasource) handleAsyncQuery(ctx context.Context, req backend.
 		fillMode = q.FillMissing
 	}
 
+	asyncDB, _ := ds.driver.GetAsyncDB(ds.connSettings, q.ConnectionArgs)
 	if q.QueryID == "" {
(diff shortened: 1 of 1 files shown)


iwysiu (approved):

idastambuk (approved):

fridgepoet:
I notice that near line 146, `GetDBFromQuery` in sqlds itself has the connection args that you originally linked in sqlds. What is the difference between the `db` there and the `asyncDB` from GetAsyncDB?

kevinwcyu:
> I notice that near line 146, `GetDBFromQuery` in sqlds itself has the connection args that you originally linked in sqlds. What is the difference between the `db` there and the `asyncDB` from GetAsyncDB?

@fridgepoet, the db is an actual sql connection i believe. whereas the asyncDB provides the methods for the async flow, i.e. starting a query, getting its status, etc.

PR #73, opened 2023-03-01
Fix connections for multiple async datasources

### LINKED ISSUE #72: Bug: Multiple Async Datasources overwrite each other's connections
When the AsyncAwsDatasource was added, instead of having a map of ids to connections like [sqlds](https://github.com/grafana/sqlds/blob/f35c5a62f66c02aef63b260c1e0a25568ae698b6/datasource.go#L48) it was created with a single static `AsyncDB`. This meant that if multiple async datasources were initialized, the later ones would overwrite the earlier one. Thus, all requests to AsyncDB specific functions would be made to whatever the newest async datasource was while the non-exclusive functions (like QueryContext) would be made to the correct datasource. This should be updated to a map.

iwysiu:
As described in the linked issue, the way that AsyncAWSDatasource is currently set up it overwrites the AsyncDB field with the most recently added Async data source. This changes it to a map, so that multiple async datasources can exist concurrently

Fixes #72

### DIFF
--- a/pkg/awsds/asyncDatasource.go
+++ b/pkg/awsds/asyncDatasource.go
@@ -3,6 +3,7 @@ package awsds
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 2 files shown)


kevinwcyu (approved):

idastambuk (approved):
Looks good, tested it and can't reproduce the bug anymore

fridgepoet (approved):
wow nice!

PR #85, opened 2023-07-14
Add ShouldCacheQuery util for async caching

iwysiu:
Adds a utility function so that the caching service in grafana can check whether or not the response should be cached.

The way the async queries work is that the first query is sent to the datasource and returns a queryId (this query always returns the started state unless it errors). This queryID is added to all the following queries, and when one of them returns a done response we stop querying.
Thus, we need to cache the "started" status response (which returns the id), and the "done" status response (which we need the id to get).

part of https://github.com/grafana/grafana-async-query-data-js/issues/14

### DIFF
--- a/pkg/awsds/utils_test.go
+++ b/pkg/awsds/utils_test.go
@@ -1,0 +1,61 @@
+package awsds
+
+import (
+	"testing"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/data"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestShouldCacheQuery(t *testing.T) {
+
+	testcases := []struct {
+		name        string
+		customMeta  map[string]interface{}
+		shouldCache bool
+	}{
+		{
+			"sync query should cache",
+			map[string]interface{}{"foo": "bar"},
+			true,
+		},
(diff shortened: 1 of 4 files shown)


kevinwcyu (approved):

pkg/awsds/utils_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}{
+		{
+			"sync query should cache",
+			map[string]interface{}{"foo": "bar"},
+			true,
+		},
+		{
+			"starting async query should cache",
+			map[string]interface{}{"status": "started"},
+			true,
```
fridgepoet, line 26: Just curious, what's the explanation for also caching "started" status? There is no result yet, so what is the purpose of caching this one? Noticed [here](https://github.com/grafana/grafana-async-query-data-js/issues/14#issuecomment-1607484280) that we only thought about "done" before.
(Can maybe update PR description with any explanation)

iwysiu: (i'll edit the pr description too) The way the async queries work is that the first query is sent to the datasource and returns a queryId (this query always returns the started state unless it errors). This queryID is added to all the following queries, and when one of them returns a done response we stop querying.
Thus, we need to cache the "started" status response (which returns the id), and the "done" status response (which we need the id to get).

PR #88, opened 2023-07-19
Get temporary credentials from credentials file

idastambuk:
Sets up to use shared credentials file for grafana_assume_role, as agreed, plus adding tests (thanks @fridgepoet for the help!) 🙏🏻

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -96,7 +96,6 @@ func ReadAuthSettingsFromEnvironmentVariables() *AuthSettings {
 
 // Session factory.
 // Stubbable by tests.
-//
 //nolint:gocritic
 var newSession = func(cfgs ...*aws.Config) (*session.Session, error) {
 	return session.NewSession(cfgs...)
@@ -104,13 +103,11 @@ var newSession = func(cfgs ...*aws.Config) (*session.Session, error) {
 
 // STS credentials factory.
 // Stubbable by tests.
-//
 //nolint:gocritic
 var newSTSCredentials = stscreds.NewCredentials
 
 // EC2Metadata service factory.
 // Stubbable by tests.
-//
 //nolint:gocritic
 var newEC2Metadata = ec2metadata.New
 
@@ -241,6 +238,7 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 		cfgs = append(cfgs, &aws.Config{Credentials: newRemoteCredentials(sess)})
 	case AuthTypeGrafanaAssumeRole:
 		backend.Logger.Debug("Authenticating towards AWS with Grafana Assume Role", "region", c.Settings.Region)
+		// TODO temporary profile name
 		profileName := "assume_role"
 		cfgs = append(cfgs, &aws.Config{
 			Credentials: newSharedCredentials(profileName),
@@ -265,7 +263,7 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 	expiration := time.Now().UTC().Add(duration)
 	if c.Settings.AssumeRoleARN != "" && sc.authSettings.AssumeRoleEnabled {
 		// We should assume a role in AWS
-		backend.Logger.Info("Trying to assume role in AWS", "arn", c.Settings.AssumeRoleARN)
+		backend.Logger.Debug("Trying to assume role in AWS", "arn", c.Settings.AssumeRoleARN)
 
 		cfgs := []*aws.Config{
 			{
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -116,11 +116,6 @@ var newEC2Metadata = ec2metadata.New
 var newRemoteCredentials = func(sess *session.Session) *credentials.Credentials {
 	return credentials.NewCredentials(defaults.RemoteCredProvider(*sess.Config, sess.Handlers))
 }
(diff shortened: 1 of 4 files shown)


pkg/awsds/sessions.go
```diff
@@ -119,6 +116,11 @@ var newEC2Metadata = ec2metadata.New
 var newRemoteCredentials = func(sess *session.Session) *credentials.Credentials {
 	return credentials.NewCredentials(defaults.RemoteCredProvider(*sess.Config, sess.Handlers))
 }
+// Shared credentials factory
+// Stubbable by tests.
+var newSharedCredentials = func(profile string) *credentials.Credentials {
+	return credentials.NewSharedCredentials("", profile)
+}
```
fridgepoet, line 123: don't think you did any stubbing, so you can remove this and just call credentials as was done in the old code.

fridgepoet (approved):

pkg/awsds/sessions.go
idastambuk: Done!

fridgepoet (approved):

pkg/awsds/sessions.go
```diff
@@ -236,6 +233,11 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 		cfgs = append(cfgs, &aws.Config{Credentials: newRemoteCredentials(sess)})
 	case AuthTypeGrafanaAssumeRole:
 		backend.Logger.Debug("Authenticating towards AWS with Grafana Assume Role", "region", c.Settings.Region)
+		// TODO temporary profile name
+		profileName := "assume_role"
```
sarahzinger, line 237: maybe we can add a little checkbox on the epic for this to double check the profile name is set properly. 

Also maybe this and the credentials path should be consts at the top of the file?

pkg/awsds/sessions_test.go
```diff
@@ -61,8 +61,8 @@ func TestNewSession_AssumeRole(t *testing.T) {
 		settings := AWSDatasourceSettings{
 			AssumeRoleARN: roleARN,
 		}
-		os.Setenv(AllowedAuthProvidersEnvVarKeyName, "default")
-		os.Setenv(AssumeRoleEnabledEnvVarKeyName, "true")
+		require.NoError(t, os.Setenv(AllowedAuthProvidersEnvVarKeyName, "default"))
```
sarahzinger, line 64: just curious what this does, is it checking that setenv doesn't error out? seems maybe overly cautious no? Or is there an expected error that could happen?

pkg/awsds/sessions_test.go
```diff
@@ -254,6 +254,69 @@ func TestNewSession_AllowedAuthProviders(t *testing.T) {
...
+				RoleARN: roleARN,
+			}
+			for _, o := range options {
+				o(p)
+			}
+			require.NotNil(t, p.ExternalID)
+
+			assert.Equal(t, "grafanamagic", *p.ExternalID)
+			return credentials.NewCredentials(p)
+		}
```
sarahzinger, line 280: hey i think i must be missing something obvious but when I read through this test I don't really see where we use newSTSCredentials. It seems like we make a function, but then we never call that function. I see we do this pattern in other tests so I'm assuming I'm missing something? Like does it overwrite something being used in getSession somehow?

sarahzinger (approved):
just asking questions for my own knowledge!

pkg/awsds/sessions_test.go
fridgepoet: I think it's unlikely an error will happen, but it would be nice to know if it does instead of silently continuing in any case.

pkg/awsds/sessions_test.go
fridgepoet: Yup it overwrites what's being used [here](https://github.com/grafana/grafana-aws-sdk/blob/b55ca30a3f84aff3840cf352c6f46c46098f1cfa/pkg/awsds/sessions.go#L260). 
The overwriting in the test just happens [here](https://github.com/grafana/grafana-aws-sdk/pull/88/files/95ed7bb4a4f5097d47b39eb30f05c36d2b6679a9#diff-dd703c9579c364b829933a94e4fdfa85fcd77e7a27d1d7168e1090f638422191R268) and it's cleaned up [here](https://github.com/grafana/grafana-aws-sdk/pull/88/files/95ed7bb4a4f5097d47b39eb30f05c36d2b6679a9#diff-dd703c9579c364b829933a94e4fdfa85fcd77e7a27d1d7168e1090f638422191R264).

(For what it's worth: `newSTSCredentials` is a global variable, it was stubbed, and it effectively hides what GetSession is depending on. We could instead use dependency injection (i.e. The dependency is directly attached to the thing that depends on it. The method receiver could contain the dependency of the signature of this function could either pass in the dependency as an argument.). In my opinion stubbing has made both the production code and test code confusing to read/understand and no explicit -- (sorry but) I think it's a bit of an anti-pattern... )

idastambuk: Just to add, @fridgepoet has already taken a look at if/how we could replace the stubs in [this branch](https://github.com/grafana/grafana-aws-sdk/compare/temp-creds/shared...remove-stub-wip) and we might start replacing the tests soon. But we agreed it's probably not worth it in this PR

fridgepoet: (not sure if it's injected in the right places or if things are "proper" and didn't test manually to see if it all works)

PR #99, opened 2023-09-15
Add Amazon Managed Grafana check in the UserAgent string header

hwandersman:
To support this issue in the AWS IoT TwinMaker plugin repo: https://github.com/grafana/grafana-iot-twinmaker-app/issues/217

We want to track which of our customers are using Amazon Managed Grafana (AMG) to run this plugin, or a self-managed Grafana server. We can do this by checking the AWS_TRAVERSAL_FETCH variable. This is a configuration option for the AWS Datasource provisioner app, which is only used in AMG.

Adding this check for AMG in each AWS call will allow all AWS services that own Grafana plugins to publish metrics about Grafana and AMG usage. This is particularly important for AWS IoT TwinMaker.

The easiest way to notify AWS IoT TwinMaker is through the [user agent header](https://github.com/grafana/grafana-iot-twinmaker-app/blob/main/pkg/plugin/twinmaker/client.go#L555) in each API call. We already provide information about the Grafana build and runtime version. We should also add whether this is in AMG or not.
* Follow up will remove custom UserAgent logic in the AWS IoT TwinMaker plugin and rely on the grafana-aws-sdk to set the header

### DIFF
--- a/pkg/awsds/utils.go
+++ b/pkg/awsds/utils.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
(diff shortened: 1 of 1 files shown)


pkg/awsds/utils.go
```diff
@@ -64,15 +65,19 @@ func GetUserAgentString(name string) string {
 		grafanaVersion = "?"
 	}
 
-	return fmt.Sprintf("%s/%s (%s; %s;) %s/%s-%s Grafana/%s",
+	// Determine if running in an Amazon Managed Grafana environment
+	_, amgEnv := os.LookupEnv("AMAZON_MANAGED_GRAFANA")
```
iwysiu, line 69: I'm fine with this, but would it make sense to do what the version getenv is doing on lines 62-65? What does this add to the header if it doesn't find the environment variable?

hwandersman: We don't really care what the value is, just that the environment variable exists. If it doesn't exist and we look up the value we will have to handle the case it is empty/undefined. It's easier to just check for the variable existence.

iwysiu (approved):

PR #104, opened 2023-10-24
Adding support for temporary credentials to sigv4 workflows

sarahzinger:
The primary motivation here is to support Temporary Credentials in Opensearch.

There's some additional work we'll have to do in opensearch before that is fully possible but this is a good step of the way there. 

To test locally:
- pull down https://github.com/grafana/grafana-aws-sdk-react/pull/66 yarn watch, then yarn link and link in opensearch
- pull down this branch and add `replace github.com/grafana/grafana-aws-sdk => /Users/sarahzinger/code/grafana/grafana-aws-sdk` to go.mod of opensearch
- (grafana internal only) see provisioning instructions here: https://github.com/grafana/plugin-provisioning/pull/137, take note of comments section with instructions for custom.ini file and .aws/credentials file
- run mage in opensearch
- make run in main grafana
- set up a new datasource as described ^ 
- save + test is expected to fail because we do not use the same backend flow yet for the health check, so for now ignore
- go to explore and select datasource, metrics should fetch successfully

This really needs tests but I found it difficult to do so. Maybe someone wants to pair with me on it? My attempts lead me to try to a larger refactor project: https://github.com/grafana/grafana-aws-sdk/pull/103 but it felt like scope creep.

### DIFF
--- a/pkg/sigv4/sigv4.go
+++ b/pkg/sigv4/sigv4.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httputil"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


kevinwcyu (approved):

idastambuk (approved):

sarahzinger:
closing as stale, although I think this would work, we can revist if and when we need this. It was mostly a proof of concept.

PR #107, opened 2023-11-07
All feature requests open as issues rather than discussions

sarahzinger:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,8 +1,5 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
```
kevinwcyu, line 1: what's the difference between this repo and the others to have this set to true?

kevinwcyu (approved):

.github/ISSUE_TEMPLATE/config.yml
sarahzinger: I guess I sort of think of this repo as a bit more internal than others so we would maybe want to let us create new blank issues easily? But I'm not convinced this does much either way. I can put it back just in case

fridgepoet (approved):

PR #108, opened 2023-12-15
Don't use the set endpoint to get the assume role credentials

iwysiu:
In https://github.com/grafana/support-escalations/issues/8225 and https://github.com/grafana/athena-datasource/issues/296 a bug was reported regarding authentication with an endpoint and assume role. The problem was that we were setting the endpoint for the Athena service on the session that we were passing to the STS service to fetch the credentials.

Under the assumption that a user who wants to use a FIPS endpoint for a service would want to use a FIPS endpoint to get the assume role credentials, I added a function to check whether the endpoint is a FIPS endpoint and return the correct endpoint. I'm not sure that this is correct to do since I was reading that there are no real rules for how AWS endpoints are named (though from what I can tell, all the [fips endpoints we'd expect](https://aws.amazon.com/compliance/fips/#FIPS_Endpoints_by_Service) would be caught by this), and I'm not sure if the users actually care if the credentials are fetched with FIPS or if that just applies to their data.

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -255,15 +255,6 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 		panic(fmt.Sprintf("Unrecognized authType: %d", c.Settings.AuthType))
 	}
 
-	if c.Settings.Endpoint != "" {
-		cfgs = append(cfgs, &aws.Config{Endpoint: aws.String(c.Settings.Endpoint)})
-	}
-
-	sess, err := newSession(cfgs...)
-	if err != nil {
-		return nil, err
-	}
-
 	duration := stscreds.DefaultDuration
 	if sc.authSettings.SessionDuration != nil {
 		duration = *sc.authSettings.SessionDuration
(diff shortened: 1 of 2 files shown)


sarahzinger (approved):

pkg/awsds/sessions.go
```diff
@@ -319,3 +324,30 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 
 	return sess, nil
 }
+
+// getSTSEndpoint checks if the set endpoint is a fips endpoint, and if so, returns the STS fips endpoint for the same region
+func getSTSEndpoint(endpoint string) string {
```
sarahzinger, line 329: no idea if this is helpful but there is an Endpoints package in aws-sdk-go I wonder if it has something for this? https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/

idastambuk (approved):

pkg/awsds/sessions.go
iwysiu: Unfortunately they don't have a helper for checking if something is a fips endpoint 😞 I think I checked and there wasn't a helper for gettting the service from a string either.

PR #115, opened 2024-01-17
Update to sqlutil helpers

njvrzm:
This updates the sdk to bring it in sync with changes to `sqlds` and `grafana-plugin-sdk-go` related to https://github.com/grafana/sqlds/issues/85 and https://github.com/grafana/grafana-plugin-sdk-go/issues/858.

Depends on https://github.com/grafana/sqlds/pull/105, https://github.com/grafana/grafana-plugin-sdk-go/pull/862 and https://github.com/grafana/grafana-plugin-sdk-go/pull/863.


sarahzinger (approved):
Seems reasonable to me! I long for the day when grafana-aws-sdk doesn't know what sqlds or sqlutils are, but today is not that day, and in the meantime it makes sense to me that we'd want it all referencing one place!

pkg/sql/api/api.go
iwysiu, line 1: It may also make sense to add `aws.Context` to the arguments for the `SQL.Stop` function, since it's also a request.

go.mod
```diff
@@ -4,18 +4,22 @@ go 1.21
 
 toolchain go1.21.3
 
+// TODO: remove both of the following before release
```
iwysiu, line 7: Reminder to remove these when you merge

iwysiu (approved):
Nice! Also thanks for handling the v3 context update

pkg/sql/api/api.go
njvrzm: Makes sense, but I think I'll leave that for a separate PR.

PR #121, opened 2024-01-30
Deprecate using environment variables for auth settings in sessions

iwysiu:
Previously, GetSessions would use the environment variables to get the grafana aws settings. This updates it to use settings passed in from the datasource and fallback to using the environment variables if they aren't set.

I also updated the changelog so that doesn't have to be a separate commit

for https://github.com/grafana/grafana/pull/81208

### DIFF
--- a/pkg/awsds/authSettings.go
+++ b/pkg/awsds/authSettings.go
@@ -91,6 +91,15 @@ func ReadAuthSettingsFromContext(ctx context.Context) (*AuthSettings, bool) {
 		hasSettings = true
 	}
 
+	if v := cfg.Get(SessionDurationEnvVarKeyName); v != "" {
+		sessionDuration, err := gtime.ParseDuration(v)
+		if err == nil {
+			settings.SessionDuration = &sessionDuration
+		} else {
+			backend.Logger.Error("could not parse env variable", "var", SessionDurationEnvVarKeyName)
+		}
+	}
+
 	if v := cfg.Get(GrafanaListMetricsPageLimit); v != "" {
 		listMetricsPageLimit, err := strconv.Atoi(v)
 		if err == nil {
@@ -109,17 +118,6 @@ func ReadAuthSettingsFromContext(ctx context.Context) (*AuthSettings, bool) {
 			backend.Logger.Error("could not parse context variable", "var", proxy.PluginSecureSocksProxyEnabled)
 		}
 		hasSettings = true
-	}
-
-	// Users set session duration directly as an environment variable
-	sessionDurationString := os.Getenv(SessionDurationEnvVarKeyName)
-	if sessionDurationString != "" {
-		sessionDuration, err := gtime.ParseDuration(sessionDurationString)
-		if err == nil {
-			settings.SessionDuration = &sessionDuration
-		} else {
-			backend.Logger.Error("could not parse env variable", "var", SessionDurationEnvVarKeyName)
-		}
 	}
 
 	return settings, hasSettings
--- a/pkg/awsds/authSettings.go
+++ b/pkg/awsds/authSettings.go
@@ -25,8 +25,8 @@ const (
 	// GrafanaAssumeRoleExternalIdKeyName is the string literal for the grafana assume role external id environment variable key name
 	GrafanaAssumeRoleExternalIdKeyName = "AWS_AUTH_EXTERNAL_ID"
 
-	// GrafanaListMetricsPageLimit is the string literal for the cloudwatch list metrics page limit key name
-	GrafanaListMetricsPageLimit = "AWS_CW_LIST_METRICS_PAGE_LIMIT"
+	// ListMetricsPageLimitKeyName is the string literal for the cloudwatch list metrics page limit key name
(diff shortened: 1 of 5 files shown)


pkg/awsds/sessions.go
```diff
@@ -106,8 +105,12 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 		// DefaultRegion is deprecated, Region should be used instead
 		c.Settings.Region = c.Settings.DefaultRegion
 	}
+	if c.AuthSettings == nil {
```
iwysiu, line 108: If we're getting them from contexts, the values will already be on the settings. Otherwise, we'll need to get them from the env variables

pkg/awsds/authSettings.go
```diff
@@ -91,6 +91,15 @@ func ReadAuthSettingsFromContext(ctx context.Context) (*AuthSettings, bool) {
 		hasSettings = true
 	}
 
+	if v := cfg.Get(SessionDurationEnvVarKeyName); v != "" {
```
sarahzinger, line 94: Should the const still be called `SessionsDurationEnvVarKeyName`?

pkg/awsds/sessions.go
sarahzinger: Maybe this comment should be in the code?

pkg/awsds/sessions_test.go
```diff
@@ -103,45 +109,89 @@ func TestNewSession_AssumeRole(t *testing.T) {
...
-		require.NoError(t, os.Setenv(AssumeRoleEnabledEnvVarKeyName, "true"))
-		require.NoError(t, os.Setenv(SessionDurationEnvVarKeyName, "20m"))
-		cache := NewSessionCache()
-		sess, err := cache.GetSession(SessionConfig{Settings: settings})
-		require.NoError(t, err)
-		require.NotNil(t, sess)
-		expCreds := credentials.NewCredentials(&stscreds.AssumeRoleProvider{
-			RoleARN:  roleARN,
-			Duration: 1200000000000, //20 minutes in nanoseconds count
+		var expectedDuration time.Duration = 1200000000000 //20 minutes in nanoseconds count
```
njvrzm, line 116: same tiny nit here

njvrzm (approved):
Looks good! Only the tiniest of nits in comments.

pkg/awsds/sessions.go
```diff
@@ -106,8 +105,12 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
...
+		c.AuthSettings = ReadAuthSettingsFromEnvironmentVariables()
+	}
+
 	authTypeAllowed := false
-	for _, provider := range sc.authSettings.AllowedAuthProviders {
```
sarahzinger, line 110: this change of going over the config vs whats in the session cache seems like it could have unexpected changes? Is it related to this pr? Sorry if I missed it!

sarahzinger (approved):
Is there a general plan to someday remove support for this entirely? Maybe we need to make a ticket and put it in out backlog to remove this from the env and then link that throughout the code so we know when we can remove this?

pkg/awsds/sessions.go
iwysiu: It's related! To stop relying on the environment variables, we need to get the GrafanaConfig off the context, and we don't have access to the context at when we create the SessionCache in the datasource. Instead, the datasource gets the config off the context when we create a new instance, and if it doesn't it gets the config off the environment variables (what currently happens)
The only unexpected thing that I can think of is if they were setting ExternalID (which is experimental and grafana only) or SessionDuration (which isn't a documented and is basically an old experimental feature) through env variables, in which case those will be ignored. Everything else gets overwritten with defaults when Grafana starts up if they weren't set in the ini file.

pkg/awsds/authSettings.go
```diff
@@ -25,8 +25,8 @@ const (
 	// GrafanaAssumeRoleExternalIdKeyName is the string literal for the grafana assume role external id environment variable key name
 	GrafanaAssumeRoleExternalIdKeyName = "AWS_AUTH_EXTERNAL_ID"
 
-	// GrafanaListMetricsPageLimit is the string literal for the cloudwatch list metrics page limit key name
-	GrafanaListMetricsPageLimit = "AWS_CW_LIST_METRICS_PAGE_LIMIT"
+	// ListMetricsPageLimitKeyName is the string literal for the cloudwatch list metrics page limit key name
+	ListMetricsPageLimitKeyName = "AWS_CW_LIST_METRICS_PAGE_LIMIT"
```
iwysiu, line 29: Discussing the other variable names made me realize that I wanted to rename this one to be more in line with the others before anyone's using it

iwysiu (commented):
> Is there a general plan to someday remove support for this entirely? Maybe we need to make a ticket and put it in out backlog to remove this from the env and then link that throughout the code so we know when we can remove this?

For multitenancy, we want plugins to stop relying on env variables. Yeah, made an issue https://github.com/grafana/grafana-aws-sdk/issues/122

PR #124, opened 2024-01-31
Add health check for async queries

sarahzinger:
[We don't have proper health checks it seems for Athena](https://github.com/grafana/athena-datasource/issues/305). this seems to fix it for me. Arguably its very weird to add to this lib, (should it be in sqlds? should it be implemented by each lib?) My thinking is this was easy and if QueryData is going to live here, health check probably should too? At least until we find another path for it? 

I have a feeling we can re-architect a lot of this when we move to the new aws-sdkv2, so hoping this is a temp fix.

Fixes: https://github.com/grafana/athena-datasource/issues/305

### DIFF
--- a/pkg/awsds/asyncDatasource.go
+++ b/pkg/awsds/asyncDatasource.go
@@ -131,25 +131,20 @@ func (ds *AsyncAWSDatasource) QueryData(ctx context.Context, req *backend.QueryD
 }
 
 func (ds *AsyncAWSDatasource) CheckHealth(ctx context.Context, req *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
-	res, err := ds.QueryData(ctx, &backend.QueryDataRequest{
-		PluginContext: req.PluginContext,
-		Queries: []backend.DataQuery{
-			{
-				RefID: "health",
-				JSON:  []byte(`{"rawSql":"SELECT 1"}`),
-			},
-		},
-	})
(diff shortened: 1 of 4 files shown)


pkg/awsds/asyncDatasource.go
```diff
@@ -130,6 +130,34 @@ func (ds *AsyncAWSDatasource) QueryData(ctx context.Context, req *backend.QueryD
...
+			Status:  backend.HealthStatusError,
+			Message: res.Responses["health"].Error.Error(),
+		}, nil
+	}
+	return &backend.CheckHealthResult{
+		Status:  backend.HealthStatusOk,
+		Message: "Data source is working",
+	}, nil
+}
+
```
kevinwcyu, line 160: We should be calling `Ping` which we already have implemented in Athena.

https://github.com/grafana/athena-datasource/blob/5debf97e96ad4148a0c6793a0295a8e8f2506c13/pkg/athena/driver/connection.go#L71-L77

sqlds implementation for how it's called

https://github.com/grafana/sqlds/blob/71090eb4d9b6af7c745e2c238131e8d107e63191/datasource.go#L311-L323

I'm not sure what the correct way to implement it here though, the awsds -> sqlds stuff is still confusing to me. But the datasources should implement the `Ping` method for us to call so they can decide how they want to check the health of the service.

sarahzinger: Cool good idea! Didn't notice we already implemented it!

kevinwcyu (approved):

PR #128, opened 2024-02-16
Use sqlds updated-time-based cache keys

njvrzm:
This updates the aws sdk to use the new caching mechanism in sqlds that includes the dataource updated timestamp in the cache key, to address https://github.com/grafana/athena-datasource/issues/264. Any other awsds- or sqlds-based could be affected by the same problem.

See the associated PR in sqlds (https://github.com/grafana/sqlds/111) for more details; this PR should not be merged until that one is released and the replace directive in `go.mod` is removed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,9 @@ go 1.21
 
 toolchain go1.21.3
 
+// FIXME: remove this when the sqlds change is released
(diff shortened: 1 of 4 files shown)


njvrzm:
I've closed the sqlds PR - that fix worked but was unnecessarily deep. The problem is confined to aws datasources (ab)using `AsyncAWSDatasource.NewDatasource`. Closing this one too.

PR #129, opened 2024-02-23
Sessions: Use STS regional endpoint in assume role for opt-in regions

idastambuk:
A [bug](https://github.com/grafana/support-escalations/issues/9155) was reported in cloudwatch when requesting resources and querying data for opt-in regions when the account is assuming a role. An error is returned from these endpoints `InvalidClientTokenId: The security token included in the request is invalid`

Opt-in regions are regions in AWS that have to be manually enabled in an AWS account. 

When requesting tokens with Amazon STS like we do when assuming role, we have to pass the `STSRegionalEndpoint` field with value 'regional' so that the returned tokens can be used in opt in regions. Otherwise global endpoint is used, which returns tokens that don't work with opt-in regions (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html)

Screenshot 2024-02-22 at 18 19 25

[As of aws-sdk-go-v2](https://docs.aws.amazon.com/sdkref/latest/guide/feature-sts-regionalized-endpoints.html), STS service uses regional endpoints by default so this shouldn't be high impact. There's also the best practices outlined in the docs above for v1 that suggest to use regional sts endpoints. 

So far, we've suggested users set the  `AWS_STS_REGIONAL_ENDPOINTS=regional`  option in their env vars, which aws/aws-sdk would read. However, this isn't possible for hosted Grafana instances. 

Fixes https://github.com/grafana/support-escalations/issues/9155

To test this, you can use the assume role same account datasource from plugin provisioning, and query region eu-south-1 (opt in region that was manually enabled in the account). You should be getting "invalid token" before the fix, and query and resource requests with empty data after it and no errors

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -170,7 +170,11 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 			// When assuming a role, the real region is set later in a new session
 			// so we use a well-known region here (not opt-in) to obtain valid credentials
 			regionCfg = &aws.Config{Region: aws.String("us-east-1")}
-			cfgs = append(cfgs, regionCfg)
+
+			// set regional endpoint flag to obtain credentials that can be used in opt-in regions as well
+			optInRegionCfg := &aws.Config{STSRegionalEndpoint: endpoints.RegionalSTSEndpoint}
+
+			cfgs = append(cfgs, regionCfg, optInRegionCfg)
 		} else {
 			regionCfg = &aws.Config{Region: aws.String(c.Settings.Region)}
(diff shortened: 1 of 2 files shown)


iwysiu (approved):
This makes sense to me, but what region/credentials were you using to test this? (I want to try it out)

idastambuk:
> This makes sense to me, but what region/credentials were you using to test this? (I want to try it out)



> This makes sense to me, but what region/credentials were you using to test this? (I want to try it out)

Sorry about that, added testing instructions at the end!

iwysiu:
Cool! I tried it and it worked

PR #132, opened 2024-03-14
Move SigV4 config and middleware here

andresmgot:
This PR does two things:

 - Adds a helper function to read SigV4 config from context.
 - Move the SigV4 middleware code from grafana/grafana here. The related PR is here: https://github.com/grafana/grafana/pull/84462

### DIFF
--- a/pkg/sigv4/sigv4_middleware.go
+++ b/pkg/sigv4/sigv4_middleware.go
@@ -1,0 +1,46 @@
+package sigv4
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend/httpclient"
+)
+
+// SigV4MiddlewareName the middleware name used by SigV4Middleware.
+const SigV4MiddlewareName = "sigv4"
+
+var newSigV4Func = New
+
+// SigV4Middleware applies AWS Signature Version 4 request signing for the outgoing request.
+func SigV4Middleware(verboseLogging bool) httpclient.Middleware {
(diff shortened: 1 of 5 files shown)


itsmylife (approved):
LGTM

pkg/sigv4/sigv4_middleware.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend/httpclient"
+)
+
+// SigV4MiddlewareName the middleware name used by SigV4Middleware.
+const SigV4MiddlewareName = "sigv4"
+
+var newSigV4Func = New
```
sarahzinger, line 13: So we have one middleware calling another middleware now... https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/sigv4/sigv4.go#L80-L82

Does this second middeware do anything that the first middleware doesn't do? Maybe better logging and something like that? Or is this just the format that the prometheus datasource expects?

sarahzinger (approved):
Seems good to consolidate!

pkg/sigv4/sigv4_middleware.go
andresmgot: yeah, this middleware is just a wrapper around the other middleware. Its goal is to do the transformation from `httpclient.Options` -> `sigv4.Options`. It could probably be refactored and only have one but it would imply changing the signature of that `New` method (so a breaking change).

mmandrus (approved):
lgtm

PR #134, opened 2024-04-16
upgrade all deps

tristanburgess:
```
govulncheck ./...
Scanning your code and 581 packages across 96 dependent modules for known vulnerabilities...

=== Symbol Results ===

No vulnerabilities found.
```


.drone.yml
```diff
@@ -25,6 +25,13 @@ steps:
...
+  - name: vuln check
+    image: golang:1.22
+    depends_on: [clone]
+    commands:
+      - go install golang.org/x/vuln/cmd/govulncheck@latest
+      - govulncheck ./...
+
 ---
 kind: signature
 hmac: dabf2613950066fd502238aeacf42600215e93853dc525e055d32027090867db
```
tristanburgess, line 37: I'm not sure how to update this myself/if I'm supposed to

```
$ mage drone
Error: you must provide your Drone access token.
```

njvrzm: I've gone ahead and signed the change. For future reference, if you log into drone and navigate to https://drone.grafana.net/account you can get your token, then add it and the server url to your environment. After that `mage drone` should work.

tristanburgess: Thank you for doing that and pointing out how to do it in the future!

sarahzinger (approved):
ok I tried installing this into athena this morning to test it out and ran into a build error so I upgraded some more dependencies and removed the bit about getkin and it seems to work for me now. I pushed up what I have maybe @njvrzm you can give it a final look over?

njvrzm (approved):
Looks fine 👍🏻  but for the record this does break at least one plugin build (opensearch) so we'll want to be thoughtful about updating them.

PR #136, opened 2024-05-02
fix: aws sts assume role with custom endpoint

wywywywy:
PR #108 added some logic around AWS STS assume role, especially when specifying a FIPS endpoint. However, if the endpoint specified is not a FIPS endpoint, it breaks STS assume role. The was deployed from Grafana 10.4.0.

This PR fixes the issue and also adds another test case to prevent it from happening again.

Related issue with more information: https://github.com/grafana/grafana/issues/86496

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -305,5 +305,5 @@ func getSTSEndpoint(endpoint string) string {
 	if strings.Contains(endpoint, "us-gov-west-1") {
 		return "sts.us-gov-west-1.amazonaws.com"
 	}
-	return ""
+	return endpoint
 }
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -218,14 +218,18 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 	expiration := time.Now().UTC().Add(duration)
 
 	if c.Settings.Endpoint != "" {
-		var endpoint = aws.String(getSTSEndpoint(c.Settings.Endpoint))
-		backend.Logger.Debug("Using custom STS endpoint", "endpoint", endpoint)
-		cfgs = append(cfgs, &aws.Config{Endpoint: endpoint})
+		cfgs = append(cfgs, &aws.Config{Endpoint: aws.String(c.Settings.Endpoint)})
 	}
 
 	if c.Settings.AssumeRoleARN != "" && c.AuthSettings.AssumeRoleEnabled {
 		// We should assume a role in AWS
 		backend.Logger.Debug("Trying to assume role in AWS", "arn", c.Settings.AssumeRoleARN)
+
+		// If a FIPS endpoint is set, we need to use the FIPS STS endpoint
+		if c.Settings.Endpoint != "" {
+			var endpoint = aws.String(getSTSEndpoint(c.Settings.Endpoint))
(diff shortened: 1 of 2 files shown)


iwysiu (commented):
Hi @wywywywy ! Thanks for the PR, I'm sorry we didn't respond to your issue, I'll check and see why we didn't get updated about it.

I tried it locally with sts credentials and I think you also need to move the `if c.Settings.Endpoint != "" {...}` part in `sessions.go` lines 223-225 to above the assume role check (around line 218 in the same file) for this and then it'll work! (I think in the case of a non-fips endpoint we want to not use the configured endpoint for the final result)

wywywywy:
> need to move the `if c.Settings.Endpoint != "" {...}` part in `sessions.go` lines 223-225 to above the assume role check (around line 218 in the same file)

I think line 257-259 will cover that.

iwysiu:
> > need to move the `if c.Settings.Endpoint != "" {...}` part in `sessions.go` lines 223-225 to above the assume role check (around line 218 in the same file)
> 
> I think line 257-259 will cover that.

Ah, I forgot to mention it, when I tested it I needed to move that first part up and then delete lines 257-259, since when we use an sts endpoint we don't actually want to set the sts endpoint on the final session.

wywywywy:
> We could change lines 257-259 to only happen for fips or gov endpoints and I think that would handle both cases?

That fails the "Assume role is enabled with a fips endpoint" test case. Any idea why this test case is providing `athena-fips.us-east-1.amazonaws.com` while checking for `sts-fips.us-east-1.amazonaws.com`?

wywywywy:
Ok I think I got it now. Please review the latest commit.

iwysiu:
> > We could change lines 257-259 to only happen for fips or gov endpoints and I think that would handle both cases?
> 
> That fails the "Assume role is enabled with a fips endpoint" test case. Any idea why this test case is providing `athena-fips.us-east-1.amazonaws.com` while checking for `sts-fips.us-east-1.amazonaws.com`?

Yeah, I made an edit to my last comment, but the endpoint field in the config conflates 2 concepts 
- the endpoint for connecting to the STS service to get the credentials
- the endpoint for actually querying the datasource

While we probably should set those separately, that is how the auth currently works, and we already have a project to refactor it (but we should fix the bug for now).

The user for the FIPS endpoint fix wanted to set the endpoint for actually querying the datasource to `athena-fips.us-east-1.amazonaws.com`, which required us to solve for the STS endpoint since they couldn't set it (hence getting the FIPS endpoint). So we do want to set it at the end to `athena-fips.us-east-1.amazonaws.com` because the final session endpoint should be for the service we're querying. If you're configuring a normal STS endpoint (like your example), we only want to use it to connect to the STS service and not to connect to the cloudwatch service, which is what we did before PR #108.

If you're comfortable with it I can add a commit that handles the weird FIPS nuance? Basically just adding a check if the configured endpoint is a FIPS/gov endpoint and only setting it on line 260 if it is.

wywywywy:
> the endpoint field in the config conflates 2 concepts

Good to know thank you. It's not obvious.

> If you're comfortable with it I can add a commit that handles the weird FIPS nuance? 

Please do. FIPS is not something we have experience with, being outside of the US.

iwysiu:
@wywywywy , Thanks again for the issue/pr! I'll get one of my coworkers to review it tomorrow and get the pr moved through and merged.

sarahzinger (approved):

PR #139, opened 2024-05-21
Add more context handling

njvrzm:
For compatibility with [grafana-plugin-sdk](https://github.com/grafana/grafana-plugin-sdk-go) (specifically changes in [v0.177.0](https://github.com/grafana/grafana-plugin-sdk-go/releases/tag/v0.177.0) and [v0.182.0](https://github.com/grafana/grafana-plugin-sdk-go/releases/tag/v0.182.0)) we're adding `context.Context` to more places. This also brings in an interface that two client datasources ([athena](https://github.com/grafana/athena-datasource/blob/55dc027435e589950dac5d4f5f0e30bb224aaf80/pkg/athena/datasource.go#L30) and [redshift](https://github.com/grafana/redshift-datasource/blob/59d8de9f733d2889ec963cb87dafe1e129db77bf/pkg/redshift/datasource.go#L37)) have been defining identically for themselves, and streamlines the various loaders that were being passed around into a single interface that `datasource.New` will expect. Finally, `context.TODO()` is replaced with `context.Background()` in tests.

This is a breaking change, so the next release should be >= v0.26.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -13,17 +13,17 @@ steps:
   - name: build
     image: grafana/grafana-plugin-ci:1.9.5
     commands:
-      - mage -v build
+      - mage --keep -v build
 
   - name: lint
     image: grafana/grafana-plugin-ci:1.9.5
     commands:
-      - mage -v lint
+      - mage --keep -v lint
 
   - name: test
     image: grafana/grafana-plugin-ci:1.9.5
     commands:
-      - mage -v test
+      - mage --keep -v test
 
   - name: vuln check
     image: golang:1.22
--- a/pkg/sql/datasource/datasource.go
+++ b/pkg/sql/datasource/datasource.go
@@ -1,6 +1,7 @@
 package datasource
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"sync"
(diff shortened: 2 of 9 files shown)


.drone.yml
```diff
@@ -13,17 +13,17 @@ steps:
   - name: build
     image: grafana/grafana-plugin-ci:1.9.5
     commands:
-      - mage -v build
```
njvrzm, line 16: Some of the build steps are randomly failing, I believe because they share a docker volume and one step's `mage` process is deleting the `mage_output_file.go` before another step is done with it.

pkg/sql/datasource/datasource.go
```diff
@@ -14,28 +15,45 @@ import (
...
 )
 
-// AWSDatasource stores a cache of several instances.
+// AWSClient provides creation and caching of sessions, database connections, and API clients
+type AWSClient interface {
+	Init(config backend.DataSourceInstanceSettings)
+	GetDB(ctx context.Context, id int64, options sqlds.Options) (*sql.DB, error)
+	GetAsyncDB(ctx context.Context, id int64, options sqlds.Options) (awsds.AsyncDB, error)
+	GetAPI(ctx context.Context, id int64, options sqlds.Options) (api.AWSAPI, error)
+}
```
njvrzm, line 24: Athena and redshift were defining this interface for themselves, which seemed backwards to me; rather than update both copies with context arguments I've moved it here.

pkg/sql/datasource/datasource.go
```diff
@@ -14,28 +15,45 @@ import (
...
+
+type Loader interface {
+	LoadSettings(context.Context) models.Settings
+	LoadAPI(context.Context, *awsds.SessionCache, models.Settings) (api.AWSAPI, error)
+	LoadDriver(context.Context, api.AWSAPI) (driver.Driver, error)
+	LoadAsyncDriver(context.Context, api.AWSAPI) (asyncDriver.Driver, error)
+}
```
njvrzm, line 31: I don't love this approach but I like it better than passing around a bunch of typed functions. I'm open to alternatives :)

kevinwcyu (approved):

pkg/sql/datasource/datasource.go
njvrzm: That would work! I think you're right though that it's not really better. Gonna keep it this way since it's already done.

PR #144, opened 2024-06-04
Create GetSessionWithAuthSettings

sarahzinger:
When working on https://github.com/grafana/timestream-datasource/pull/285 we noticed that in order to not rely on env variables we have to grab auth settings from the datasource instance's context and pass them along to GetSession.

However this is easy to forget/miss. This pr will create a new method for getting session which is passed in the datasource instance's context with auth settings and sets it as appropriate.

### DIFF
--- a/pkg/awsds/sessions.go
+++ b/pkg/awsds/sessions.go
@@ -1,7 +1,6 @@
 package awsds
 
 import (
-	"context"
 	"crypto/sha256"
 	"fmt"
 	"net/http"
@@ -74,7 +73,7 @@ var newRemoteCredentials = func(sess *session.Session) *credentials.Credentials 
 	return credentials.NewCredentials(defaults.RemoteCredProvider(*sess.Config, sess.Handlers))
 }
 
-type GetSessionWithContextConfig struct {
+type GetSessionConfig struct {
 	Settings      AWSDatasourceSettings
 	HTTPClient    *http.Client
 	UserAgentName *string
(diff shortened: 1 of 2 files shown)


pkg/awsds/sessions.go
```diff
@@ -289,6 +296,18 @@ func (sc *SessionCache) GetSession(c SessionConfig) (*session.Session, error) {
 	return sess, nil
 }
 
+// When calling, be sure that the context passed is the datasource instance's context which contains auth settings
+func (sc *SessionCache) GetSessionWithContext(ctx context.Context, c GetSessionWithContextConfig) (*session.Session, error) {
```
iwysiu, line 300: I know that I'm the one that suggested this approach, but if we need to use the instance context every time we get a session, it may be better to make this something like `GetSessionWithAuthSettings(authSettings AuthSettings, c SessionConfig)` and remove the AuthSettings from the SessionConfig to make it a clearer separation (cloudwatch is the only one using `SessionConfig.AuthSettings` so it's fine). This implementation is going to end up requiring us to store the context, which is [discouraged in Go](https://go.dev/blog/context-and-structs), while if we take the `AuthSettings` as an argument we can store those on the instance instead. 

@njvrzm do you have any ideas/preferences for how to organize this?

sarahzinger:
Ok updated as you suggested @iwysiu so that auth settings are now explicitly required to be passed in (with instructions on how to do so), rather than read from the env (or pulled from a stored context)

pkg/awsds/sessions.go
```diff
@@ -73,6 +73,12 @@ var newRemoteCredentials = func(sess *session.Session) *credentials.Credentials
 	return credentials.NewCredentials(defaults.RemoteCredProvider(*sess.Config, sess.Handlers))
 }
 
+type GetSessionConfig struct {
```
iwysiu, line 76: We could probably just use SessionConfig for `GetSessionWithAuthSettings` but this works too!

iwysiu (approved):

pkg/awsds/sessions.go
sarahzinger: yeah i guess my thinking is that sessionConfig has authsettings in there which seems confusing to have a property in there we never would fill out idk.

PR #150, opened 2024-06-21
SigV4: Update to not use env variables

iwysiu:
This is to support not using environment variables in prometheus-amazon

Technically, this is a breaking change on sigv4.New. Also technically, this is an experimental library so we're allowed to break things, and it's not going to compile, so users (if there are any that aren't us) will definitely notice.

We should also update grafana with the new versions of the deprecated functions once this goes in.

### DIFF
--- a/pkg/awsds/authSettings.go
+++ b/pkg/awsds/authSettings.go
@@ -40,7 +40,7 @@ const (
 )
 
 // ReadAuthSettings gets the Grafana auth settings from the context if its available, the environment variables if not
-// This function is only for backwards compatibility, generally ReadAuthSettingsFromContext should be used instead
+// Deprecated: This function is only for backwards compatibility, generally ReadAuthSettingsFromContext should be used instead
 func ReadAuthSettings(ctx context.Context) *AuthSettings {
 	settings, exists := ReadAuthSettingsFromContext(ctx)
 	if !exists {
--- a/pkg/awsds/authSettings.go
+++ b/pkg/awsds/authSettings.go
@@ -40,6 +40,7 @@ const (
 )
 
 // ReadAuthSettings gets the Grafana auth settings from the context if its available, the environment variables if not
+// This function is only for backwards compatibility, generally ReadAuthSettingsFromContext should be used instead
 func ReadAuthSettings(ctx context.Context) *AuthSettings {
 	settings, exists := ReadAuthSettingsFromContext(ctx)
 	if !exists {
--- a/pkg/sigv4/sigv4.go
+++ b/pkg/sigv4/sigv4.go
@@ -79,6 +79,7 @@ func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 
 // New instantiates a new signing middleware with an optional succeeding
 // middleware. The http.DefaultTransport will be used if nil
+// AuthSettings can be gotten from the datasource instance's context with awsds.ReadSettingsFromContext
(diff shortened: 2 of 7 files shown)


pkg/awsds/authSettings.go
```diff
@@ -40,6 +40,7 @@ const (
 )
 
 // ReadAuthSettings gets the Grafana auth settings from the context if its available, the environment variables if not
+// This function is only for backwards compatibility, generally ReadAuthSettingsFromContext should be used instead
 func ReadAuthSettings(ctx context.Context) *AuthSettings {
```
njvrzm, line 44: I think we should probably mark this as deprecated too, since it relies on an already-deprecated function. It's only used (on all of public github, anyway) by cloudwatch, where I guess we need to maintain the legacy support. (Right?) I wonder if we should move `ReadAuthSettingsFromEnvironmentVariables` there?

pkg/sigv4/sigv4.go
```diff
@@ -79,7 +79,8 @@ func (rt RoundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 
 // New instantiates a new signing middleware with an optional succeeding
 // middleware. The http.DefaultTransport will be used if nil
-func New(cfg *Config, next http.RoundTripper, opts ...Opts) (http.RoundTripper, error) {
+// AuthSettings can be gotten from the datasource instance's context with awsds.ReadSettingsFromContext
```
njvrzm, line 82:
```suggestion
// AuthSettings can be gotten from the datasource instance's context with awsds.ReadAuthSettingsFromContext
```

pkg/sigv4/sigv4_middleware.go
```diff
@@ -12,8 +13,9 @@ const SigV4MiddlewareName = "sigv4"
 
 var newSigV4Func = New
 
-// SigV4Middleware applies AWS Signature Version 4 request signing for the outgoing request.
-func SigV4Middleware(verboseLogging bool) httpclient.Middleware {
+// SigV4MiddlewareWithAuthSettings applies AWS Signature Version 4 request signing for the outgoing request.
+// AuthSettings can be gotten from the datasource instance's context with awsds.ReadSettingsFromContext
```
njvrzm, line 17:
```suggestion
// AuthSettings can be gotten from the datasource instance's context with awsds.ReadAuthSettingsFromContext
```

pkg/sigv4/sigv4.go
njvrzm: (Might as well fix the one in `sessions.go` too: https://github.com/grafana/grafana-aws-sdk/blob/a309fdff4552e0260f10f2e67d2e81ab19803890/pkg/awsds/sessions.go#L298

njvrzm (approved):
Looks good, just a few minor suggestions.

pkg/awsds/authSettings.go
iwysiu: done! Cloudwatch is actually the one where maintaining legacy support is the least important, since it's bundled with grafana, so we should update it. TBH I was worried about the other datasources having backwards compatibility issues, but no one's complained yet.

pkg/sigv4/sigv4.go
iwysiu: done

pkg/sigv4/sigv4_middleware.go
iwysiu: done

bohandley (approved):
LGTM! 🚀

PR #151, opened 2024-07-02
Add elasticache serverless dimensions and metrics

mpjunior92:
AWS announced the availability of ElastiCache Serverless offering on their catalog back in Nov 2023 [[Amazon ElastiCache Serverless for Redis and Memcached is now available]](https://aws.amazon.com/blogs/aws/amazon-elasticache-serverless-for-redis-and-memcached-now-generally-available/). This new offering uses a new dimension named `clusterId` and added a couple of new metrics [[Serverless cache metrics](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/serverless-metrics.html)]. This PR adds support to this new dimension and these new metrics so that they are available on Grafana CloudWatch data source as well. Here is a sample of the new `ElastiCacheProcessingUnits` metric:
```json
        {
            "Namespace": "AWS/ElastiCache",
            "MetricName": "ElastiCacheProcessingUnits",
            "Dimensions": [
                {
                    "Name": "clusterId",
                    "Value": "my-redis-cluster"
                }
            ]
        },
```

Please note that these new metrics are under the same namespace, `AWS/ElastiCache`, as the traditional non-serverless metrics.


njvrzm (approved):
Looks good! Thanks for your contribution :)

pkg/cloudWatchConsts/metrics.go
```diff
@@ -523,7 +536,7 @@ var NamespaceDimensionKeysMap = map[string][]string{
 	"AWS/EFS":                            {"FileSystemId", "StorageClass"},
 	"AWS/ELB":                            {"AvailabilityZone", "LoadBalancerName"},
 	"AWS/ES":                             {"ClientId", "DomainName", "NodeId"},
-	"AWS/ElastiCache":                    {"CacheClusterId", "CacheNodeId"},
+	"AWS/ElastiCache":                    {"CacheClusterId", "CacheNodeId", "clusterId"},
```
kevinwcyu, line 539: Is `clusterId` documented somewhere, I `CTRL + F` the linked documents from the description but nothing came up.

The casing seems strange to me. If it exists, could it be `ClusterId` instead?

mpjunior92: I could not find any docs for it. Case seems a bit odd to me as well, but that is how it is defined on console:
image

mpjunior92: Same output from CLI:
image

PR #152, opened 2024-07-03
Add DatabaseCapacityUsageCountedForEvictPercentage for AWS/ElastiCache

andriikushch:
This PR adds `DatabaseCapacityUsageCountedForEvictPercentage` metric to `AWS/ElastiCache` namespace: 

Doc: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheMetrics.Redis.html


kgeckhart (approved):
While we are here, are there any other obvious metrics we might be missing?

andriikushch:
> While we are here, are there any other obvious metrics we might be missing?

I think there might be a few and I have an idea how to check it: will do it in another PR.

I will merge this PR now because we might need it for:

- https://github.com/grafana/support-escalations/issues/11237

PR #153, opened 2024-07-04
Add some missing metrics to AWS/ElastiCache

andriikushch:
This PR adds some metrics to the AWS/ElastiCache

They are collected from:

- https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheMetrics.Redis.html
- https://docs.aws.amazon.com/AmazonElastiCache/latest/mem-ug/CacheMetrics.Memcached.html
- https://docs.aws.amazon.com/AmazonElastiCache/latest/mem-ug/CacheMetrics.HostLevel.html
- https://docs.aws.amazon.com/memorydb/latest/devguide/metrics.memorydb.html

Using script:

```javascript
a = []
$x("//table/tbody/tr[count(*)=3]/td[1]/code/text()").forEach(el => a.push(el.textContent))
```

Results were aggregated, normalized, and sorted before adding to the PR.

The first commit in this PR is to sort existing metrics to make diff with the new metrics obvious.

<s>**NetworkLinkLocalAllowanceExceeded was removed** because I did not find it in the docs.</s>


cristiangreco (approved):

andriikushch:
@kgeckhart , @cristiangreco I checked the HE API code: we have validation there for the available metrics :

- https://github.com/grafana/cloud-onboarding/blob/1a710eeba07af26cf2023d6f6d6bdcd5f27cb483/pkg/hosted-exporters-api/service/service.go

Do you think it is safe to remove the metric(NetworkLinkLocalAllowanceExceeded), or shall we keep it for backward compatibility? 

I think the customer might have a problem editing the scrape job with that metric.

kevinwcyu (approved):

kgeckhart:
> **NetworkLinkLocalAllowanceExceeded was removed** because I did not find it in the docs.

We should follow up with this change as it's going to be a bit tedious to pull off without a lot of headaches.

andriikushch:
> > NetworkLinkLocalAllowanceExceeded was removed because I did not find it in the docs.

> We should follow up with this change as it's going to be a bit tedious to pull off without a lot of headaches.

@kgeckhart, I reverted `NetworkLinkLocalAllowanceExceeded` deletion

PR #155, opened 2024-07-16
Support errorsource

njvrzm:
This supports adding errorsource to clients of the sdk. One example here: https://github.com/grafana/athena-datasource/pull/344.

Also prepares for v0.29.0 release.


sarahzinger (approved):
🚀

PR #156, opened 2024-07-17
Sort NamespaceMetricsMap

andriikushch:
This PR sorts `NamespaceMetricsMap`. It will simplify adding new metrics in the future: it will be much easier to see the diff in the PRs and see which metrics were added and which were removed.

This PR includes a test to ensure that metrics are sorted and another test to ensure the correctness of the initial sorting.


kgeckhart (approved):
👏 I find it so much easier to read this file now

pkg/cloudWatchConsts/metrics_test.go
```diff
@@ -0,0 +1,618 @@
...
+	return isSorted, report
+}
+
+// todo: remove after this PR is merged
+func TestNamespaceMetricsMapIsTheSameAfterTheInitialOrdering(t *testing.T) {
+	assert.True(t, compareMapsUnordered(NamespaceMetricsMap, originalNamespaceMetricsMap))
+}
+
+// todo: remove after this PR is merged
+var originalNamespaceMetricsMap = map[string][]string{
```
kgeckhart, line 81: I think it's fine to remove this before you merge it vs merging and needing to do a follow up. There will always be the changeset which shows it is true

iwysiu (approved):
Looks good! just remember to add `AWS/ElastiCache` https://github.com/grafana/grafana-aws-sdk/pull/153 and I agree with Kyle about removing the test that compares it to the old version.

PR #158, opened 2024-07-25
Add expected casing for AWS/Kafka TCPConnections

kgeckhart:
This PR adds what looks to be the proper casing for TCP Connections for AWS/Kafka. I added it vs replacing it because I'm not quite sure what it would mean to remove the other one for the data source.

The AWS documentation shows the value we use today, 
image
https://docs.aws.amazon.com/msk/latest/developerguide/metrics-details.html

But when looking at this metric within our own AWS systems it appears to be `TCPConnections`,
image 
and because metrics are case sensitive it causes some problems with using the metric with the current casing.

### DIFF
--- a/pkg/cloudWatchConsts/metrics.go
+++ b/pkg/cloudWatchConsts/metrics.go
@@ -1668,6 +1668,7 @@ var NamespaceMetricsMap = map[string][]string{
 		"SumOffsetLag",
 		"SwapFree",
 		"SwapUsed",
+		"TCPConnections",
 		"TcpConnections",
 		"TrafficBytes",
 		"TrafficShaping",


iwysiu (commented):
What I suspect happened is that they just put the wrong metric name in the docs, and you can probably remove `TcpConnections`. If you want to check from the aws api you can use https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/list-metrics.html to get the metrics and those should be right

kgeckhart:
> What I suspect happened is that they just put the wrong metric name in the docs, and you can probably remove `TcpConnections`. If you want to check from the aws api you can use https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/list-metrics.html to get the metrics and those should be right

I removed `TcpConnections` but then decided to put it back after thinking further. The second screenshot showing it as `TCPConnections` came straight from the CloudWatch console which is how I discovered the casing issue. What I don't know is if this was always like this or if a change on AWS's side caused it at a certain version of MSK 🤔.

iwysiu (approved):
I would not be surprised if the aws docs have always been wrong and its safe to remove, but I don't think it's a big deal either way.

PR #161, opened 2024-08-07
Add more metrics to the services

andriikushch:
This PR:
- Adds more metrics to the services
- Adds new services
- Adds dimensions to the new services


kgeckhart (approved):
:shipit:

pkg/cloudWatchConsts/metrics.go
```diff
@@ -3083,4 +3204,6 @@ var NamespaceDimensionKeysMap = map[string][]string{
 	"AWS/Cassandra":                               {"Keyspace", "Operation", "TableName"},
 	"AWS/AmplifyHosting":                          {"App"},
 	"CloudWatchSynthetics":                        {"CanaryName", "StepName"},
```
iwysiu, line 3206: Even if it's empty we should add `AWS/ECR` here (I think it errors if it doesn't exist)

iwysiu (approved):
Looks good with one change!

pkg/cloudWatchConsts/metrics.go
andriikushch: Thank you @iwysiu!

Added here https://github.com/grafana/grafana-aws-sdk/pull/161/commits/692a75fa2e53a0b0c7aa422ae16d680972403aba.

PR #163, opened 2024-08-08
For the backward compatibility revert metric name deletion

andriikushch:
Revert metric name deletion from 

- https://github.com/grafana/grafana-aws-sdk/pull/161/files


andriikushch:
Not needed anymore.

PR #172, opened 2024-09-19
Update CloudWatch Metrics for AWS IoT SiteWise

Deepali1211:
Update CloudWatch Metrics for AWS IoT SiteWise.

Issue ref# https://github.com/grafana/grafana/issues/92747


iwysiu:
Hi @Deepali1211 , thanks for the pr! Based on the [AWS docs](https://docs.aws.amazon.com/iot-sitewise/latest/userguide/monitor-cloudwatch-metrics.html), it looks like the the old `Gateway` metrics were also valid, so I think we should put those back and have both the `Gateway` and the `IoTSiteWisePublisher` metrics. Also could you add the other metrics from the docs so that we have all of them?

Deepali1211:
I have added all the Version 1 and Version 2 metrics.

iwysiu (commented):
Hi @Deepali1211 , thanks for adding all the metrics!
We're almost there, 2 more changes and I can merge it
1) If you run `mage test` in the file, it'll alphabetize the `AWS/IotSiteWise` metrics for you and can copy that into the file.
2) Can you replace the `AWS/IotSiteWise` line on 3295 (in the Dimensions Map) with 

```
"AWS/IoTSiteWise":                    {"AccountId", "GatewayId", "PropertyGroup", "Reason", "SourceName", "StreamName", "ThrottledAt"},
```

Deepali1211:
Done.
Also, there are entries like "OpcUaCollector.Heartbeat" and "OPCUACollector.Heartbeat". 
Considering keys to be case insensitive, I have kept single entry.
Do you want me to keep metrics with both keys `OpcUaCollector` and `OPCUACollector`?

iwysiu:
> Done. Also, there are entries like "OpcUaCollector.Heartbeat" and "OPCUACollector.Heartbeat". Considering keys to be case insensitive, I have kept single entry. Do you want me to keep metrics with both keys `OpcUaCollector` and `OPCUACollector`?

Thanks! I think we've found in the past that metrics are case sensitive, so can you keep both keys?

Deepali1211:
Added two metrics with `OPCUACollector` as per their docs.
They only have two metrics with OPCUACollector and have duplicate entries with OpcUaCollector.
image

iwysiu (approved):
Thanks @Deepali1211 ! I'll get this merged and into the Cloudwatch plugin

PR #183, opened 2024-12-02
Add metrics for elasticache serverless

kgeckhart:
Adds some newer metrics from https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/serverless-metrics-events-redis.html and https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/serverless-metrics-events.memcached.html


njvrzm (approved):
Looks good, thank you!

PR #185, opened 2024-12-04
Add aws-sdk-go-v2 credentials provider (session wrapper)

njvrzm:
This extracts some code from the [Redshift V2 migration](https://github.com/grafana/redshift-datasource/pull/325) for use in future migrations. It should be considered experimental for now, but it does work.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ toolchain go1.23.2
 
 require (
 	github.com/aws/aws-sdk-go v1.55.5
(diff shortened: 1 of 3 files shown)


iwysiu (approved):
This looks good for getting the redshift poc in, but if I'm correct this is still dependent on the aws-sdk v1, right?

njvrzm:
Yes, definitely - it's a temporary workaround to enable us to migrate the individual datasources' direct use of the AWS SDK to v2 independely from migrating our auth functionality here.

PR #202, opened 2025-02-12
Add support for aws-sdk-go-v2 authentication

njvrzm:
This adds support for the new authentication style of aws-sdk-go-v2. It exports one method, `GetAWSConfig`, called with settings to allow all the auth mechanisms we support. All except EC2 IAM Role have been tested.


pkg/awsauth/auth.go
```diff
@@ -0,0 +1,53 @@
...
+	authType := authSettings.GetAuthType()
+	logger.Debug(fmt.Sprintf("Using auth type: %s", authType))
+	switch authType {
+	case AuthTypeDefault: // nothing else to do here
+	case AuthTypeKeys:
+		options = append(options, authSettings.WithStaticCredentials(client))
+	case AuthTypeSharedCreds, AuthTypeGrafanaAssumeRole:
+		options = append(options, authSettings.WithSharedCredentials())
+	case AuthTypeEC2IAMRole:
+		// TODO: test this
```
iwysiu, line 28: Did this get tested?

pkg/awsauth/auth.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func getAWSConfigWithClient(ctx context.Context, authSettings Settings, client AWSAPIClient) (aws.Config, error) {
+	logger := backend.Logger.FromContext(ctx)
+	options := authSettings.BaseOptions()
+
+	authType := authSettings.GetAuthType()
+	logger.Debug(fmt.Sprintf("Using auth type: %s", authType))
+	switch authType {
+	case AuthTypeDefault: // nothing else to do here
```
iwysiu, line 22: is the default keys? I think this is going to fall through to keys

iwysiu (commented):
I didn't get a chance to look super closely at this, but did you try this with both fips and non-fips sts endpoints? I know we had a weird case with the original where fips nedpoints need to be used differently than non-fips endpoints

pkg/awsauth/auth.go
njvrzm: No, the todo is really a note for later. Since this new auth approach is opt-in for now I think it's okay to come back to it.

pkg/awsauth/auth.go
njvrzm: Go cases only fall through explicitly, if you add the `fallthrough` keyword: https://go.dev/wiki/Switch#fall-through.

njvrzm:
Hm - I think we may need to rework how FIPS endpoints are handled anyway. I just tried one with prod cloudwatch and the endpoints are service-specific, so you can only either use metrics or logs but not both. I think an option for "use FIPS endpoints" would make more sense, and it would be easy to support in the new auth system. Is there any other reason for folks to use the custom endpoint feature?
 
In the mean time, I'll add a workaround to check for and remove fips from the endpoint and handle it correctly for both services (and STS). Nice catch! I didn't know about this.

iwysiu (commented):
> Hm - I think we may need to rework how FIPS endpoints are handled anyway. I just tried one with prod cloudwatch and the endpoints are service-specific, so you can only either use metrics or logs but not both. I think an option for "use FIPS endpoints" would make more sense, and it would be easy to support in the new auth system. Is there any other reason for folks to use the custom endpoint feature?
> 
> In the mean time, I'll add a workaround to check for and remove fips from the endpoint and handle it correctly for both services (and STS). Nice catch! I didn't know about this.

Yeah, I bring it up in the fips issues: https://github.com/grafana/grafana-aws-sdk/pull/108 and https://github.com/grafana/grafana-aws-sdk/pull/136 but we should honestly be accepting an endpoint for auth and an endpoint for the service. The issue is that gov users use it to specify the endpoint for the service (and I made a [helper](https://github.com/grafana/grafana-aws-sdk/blob/7d9319dabe772130fb5c81d77a173375d756de86/pkg/awsds/sessions.go#L326) to get the matching fips endpoint for sts) while the other users would use it to specify the sts endpoint and then want the normal endpoint for the service (i.e. not setting the endpoint).

iwysiu (approved):
Overall looks good! Do you have a branch of cloudwatch pushed up that uses the new logic to try?

njvrzm:
PR for cloudwatch update is here: https://github.com/grafana/grafana/pull/99643. We can give it a test during team time later.

PR #205, opened 2025-02-17
AWSDS: Add QueryExecutionError type, prepare 0.32.0

idastambuk:
The `queryData` endpoint returns HTTP Status 400 if the query failed. This is the case for downstream errors, be they user errors, timeout errors or internal server errors etc.  However, we'd like to pass on the cause of the downstream error if we know it. Presumably the field `errorResponse.status` is the place for that (500 for internal, 400 for user errors etc.). In order to do that, I've tried to define a new custom error type that can be returned from datasource.Execute or datasource.GetQueryStatus (for async datasources), which gets passed on as errorResponse.status:
Screenshot 2025-02-17 at 18 57 52
I thought this could be a good way to optionally pass the cause with the error from the datasource, and it can be extended with other error causes if necessary. But open to other suggestions if this is not a good pattern. 

Here's how it could be implemented in the datasource: https://github.com/grafana/athena-datasource/pull/506

This was prompted by [this bug report](https://github.com/grafana/athena-datasource/issues/324) where the user is using the http status from the endpoint to retry or cancel the query/ Since we can't change the HTTP status from queryData, hopefully they can read the errorResponse.status code at least. In any case this is an improvement for Athena datasource that we could use for monitoring downstream errors, since status is included in our server logs.


pkg/awsds/asyncDatasource.go
```diff
@@ -119,7 +119,19 @@ func (ds *AsyncAWSDatasource) QueryData(ctx context.Context, req *backend.QueryD
 			var err error
 			frames, err = ds.handleAsyncQuery(ctx, query, req.PluginContext.DataSourceInstanceSettings.UID)
 			if err != nil {
-				response.Set(query.RefID, errorsource.Response(err))
+				errorResponse := errorsource.Response(err)
+				// checking if we know the cause of downstream error
+				if e, ok := err.(*QueryExecutionError); ok {
```
iwysiu, line 124: I think it would be more robust to use `errors.As` here

iwysiu (approved):

pkg/awsds/asyncDatasource.go
idastambuk: Done!

PR #214, opened 2025-03-19
Fix: use alternate STS endpoint for STS interaction if given

njvrzm:
This fixes a regression from the fix in #136: if the datasource config gives an endpoint and that endpoint is for STS, we need to use that endpoint for STS interaction but not afterwards.

This also fixes a unit test from the previous release that was failing, and prepares the next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## 0.33.1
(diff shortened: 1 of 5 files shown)


idastambuk (approved):

njvrzm:
Something odd happened with this PR: I got some errors when trying to merge it, but the squashed commit was added to main. The PR now says "merge already in progress" if I try to merge it. I'll close it I guess?

PR #216, opened 2025-03-19
feat: metrics for lambda event source mappings

rrhodes:
# What?

Adding Lambda [event source mapping metrics](https://docs.aws.amazon.com/lambda/latest/dg/monitoring-metrics-types.html#event-source-mapping-metrics) to CloudWatch constants.


idastambuk (approved):
Thanks a lot for the contribution @rrhodes!

PR #226, opened 2025-03-31
Use DefaultClient if given nil HTTPClient

njvrzm:
SecureSocksProxy is currently failing with the new v2 auth; the log doesn't contain enough details to be sure but I believe it's because the HTTPClient is coming in nil.

### DIFF
--- a/pkg/awsauth/settings.go
+++ b/pkg/awsauth/settings.go
@@ -181,6 +181,9 @@ func (s Settings) WithHTTPClient() LoadOptionsFunc {
 		if s.HTTPClient != nil {
 			options.HTTPClient = s.HTTPClient
 		}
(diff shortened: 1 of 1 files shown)


idastambuk:
Tried it with njvrzm/migrate-cloudwatch-to-aws-sdk-go-v2 branch without this PR and I get `1. CloudWatch metrics query failed: cfg.HTTPClient is not *http.Client 2. CloudWatch logs query failed: cfg.HTTPClient is not *http.Client`. Then I replaced go.sum with this version and it works fine, I can see the requests in PDC server 👏🏻

https://wiki.grafana-ops.net/w/index.php/Engineering/Grafana/Data_Sources/API_servers/Testing_datasources_with_PDC_Locally

PR #229, opened 2025-04-10
Fix: clone default transport instead of using it for PDC

njvrzm:
With the current code when PDC is applied it applies to the default transport, which causes problems in mutli-tenant mode.


evandandrea (approved):
LGTM

idastambuk (approved):

PR #231, opened 2025-04-14
Add support for multitenant temporary credentials to v1 path

iwysiu:
Updates the aws-sdk-v1 path to get the access and secret key for multitenant temp creds and prepares the release.

### DIFF
--- a/pkg/awsds/types.go
+++ b/pkg/awsds/types.go
@@ -19,15 +19,15 @@ type AmazonSessionProvider func(region string, s AWSDatasourceSettings) (*sessio
 
 // AuthSettings stores the AWS settings from Grafana
 type AuthSettings struct {
-	AllowedAuthProviders []string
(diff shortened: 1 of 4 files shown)


pkg/awsds/types.go
```diff
@@ -26,7 +26,8 @@ type AuthSettings struct {
 	ListMetricsPageLimit int
 
 	// necessary for a work around until https://github.com/grafana/grafana/issues/39089 is implemented
-	SecureSocksDSProxyEnabled bool
+	SecureSocksDSProxyEnabled  bool
```
njvrzm, line 29: Might be nice to note in the comment here that it's specifically `SecureSocksProxyDSEnabled` that's a workaround, or move the new attribute above the comment

njvrzm (approved):
Looks good! One minor suggestion but 👍🏻

PR #235, opened 2025-05-01
Fix check for multitenant temporary credentials

iwysiu:
I think this should fix the file checking for instances that are not remote


njvrzm (approved):
Dig it 👍🏻

PR #242, opened 2025-05-12
Add CloudWatch AWS/EKS metrics and dimensions

jangaraj:
Source: https://docs.aws.amazon.com/eks/latest/userguide/cloudwatch.html


idastambuk:
Hi @jangaraj, thanks a lot for adding these! I guess it makes sense that it would be there, but I don't see `ClusterName` dimension in the docs, is that something that is missing on their side or extra on ours?

jangaraj:
Yes, AWS doc is not complete. I see that dimension in my CloudWatch EKS metric.

idastambuk (approved):

PR #250, opened 2025-06-04
Add missing opt in regions (thailand/mexico)

chriscerie:
Add missing opt in regions (Thailand and Mexico central) from [AWS docs](https://docs.aws.amazon.com/controltower/latest/userguide/opt-in-region-considerations.html).


njvrzm (changes requested):
Looks like we're missing one more region - `ap-east-2` is listed as opt-in in the doc you linked but isn't in our list. Would you mind adding that one too, please?

chriscerie:
Thanks @njvrzm, looks like it was added to docs within the past few days

njvrzm (approved):
:shipit:

PR #258, opened 2025-06-26
Remove aws-sdk-go v1 entirely

njvrzm:
This removes all traces of aws-sdk-go v1 so that it is not a transitive dependency for users of this library.


idastambuk (approved):
Can we also get [this](https://github.com/grafana/grafana-aws-sdk/pull/259) into grafana-aws-sdk@v1?

PR #259, opened 2025-06-27
Chore: Migrate to github actions for CI

idastambuk:
Also fixes some lint issues. 
We were getting ` error var StopError should have name of the form ErrFoo` for errors in api.go, so I changed their name. This means that this will be a breaking change for plugins when they update, but it should just be simple to fix it. Nevertheless, we should probably aim to release this together with the v1 deprecation release we're planning soon


njvrzm (approved):
sweeet

PR #283, opened 2025-07-30
Bugifx: Fix custom sts endpoint

idastambuk:
There was a bug where setting the default sts endpoint (`sts.region.amazonaws.com`) resulted in an authentication error. Also fixed the case where adding https:// prefix didn't correctly classify endpoint as sts.

### DIFF
--- a/pkg/awsauth/auth.go
+++ b/pkg/awsauth/auth.go
@@ -83,5 +83,5 @@ func (rcp *awsConfigProvider) GetConfig(ctx context.Context, authSettings Settin
 }
 
 func isStsEndpoint(ep *string) bool {
-	return ep != nil && (strings.HasPrefix(*ep, "sts.") || strings.HasPrefix(*ep, "sts-fips."))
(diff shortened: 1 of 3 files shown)


pkg/awsauth/auth.go
```diff
@@ -83,5 +83,5 @@ func (rcp *awsConfigProvider) GetConfig(ctx context.Context, authSettings Settin
 }
 
 func isStsEndpoint(ep *string) bool {
-	return ep != nil && (strings.HasPrefix(*ep, "sts.") || strings.HasPrefix(*ep, "sts-fips."))
+	return ep != nil && (strings.Contains(*ep, "sts.") || strings.Contains(*ep, "sts-fips."))
```
kevinwcyu, line 86: this could affect some endpoints that aren't sts endpoints. E.g. something like https://athenatests.example.com would be considered as an sts endpoint. We could keep a list of prefixes ["https://sts.", "https://sts-fips.", etc] and check if the endpoint prefix is one of those in the list.

kevinwcyu (approved):

PR #338, opened 2025-09-30
Add handling for new grafana assume role keys

iwysiu:
The new keys are located at a new folder and there may be multiple of them, so updating the logic for that.


njvrzm (changes requested):
One blocking comment I'm afraid - let me know if you want to chat about approaches

PR #355, opened 2025-10-24
Add missing ap-southeast-6 opt in region

chriscerie:
See https://github.com/grafana/grafana/issues/113011

Add missing ap-southeast-6 opt in region. Docs: https://docs.aws.amazon.com/general/latest/gr/cw_region.html


njvrzm (approved):
👍🏻👍🏻👍🏻

PR #474, opened 2026-07-20
fix(awsauth): mark SigV4 credential/config failures as downstream errors

dgiagio:
## What

Wrap the SigV4 signer's credential and config resolution failures in `backend.DownstreamError`.

## Why

When `GetConfig` or `cfg.Credentials.Retrieve` fails in `SignerRoundTripper.RoundTrip` (e.g. a denied `sts:AssumeRole`, or expired/invalid credentials), the error was returned unwrapped. With no error source attached, Grafana defaults it to `plugin`, so customer-side AWS auth problems get charged against the datasource plugin's error-rate SLO instead of being reported as downstream.

## Root cause analysis

https://github.com/grafana/data-sources/issues/1539


mattvella07 (approved):
LGTM

PR #481, opened 2026-08-04
Rename Avails MediaTailor metrics to Avail

### LINKED ISSUE #130115: CloudWatch: AWS/MediaTailor Avail.* metrics are misnamed Avails.* in metric suggestions
**What happened**
In the CloudWatch query editor, the metric-name suggestions for the `AWS/MediaTailor` namespace offer `Avails.Duration`, `Avails.FillRate`, and `Avails.FilledDuration`. These metrics do not exist in CloudWatch, so queries built from the suggestions return no data even when the account has MediaTailor data.

**What was expected**
AWS publishes these metrics with a singular prefix: `Avail.Duration`, `Avail.FillRate`, `Avail.FilledDuration`. See the [AWS MediaTailor CloudWatch metrics documentation](https://docs.aws.amazon.com/mediatailor/latest/ug/monitoring-cloudwatch-metrics.html).

**How to reproduce**

1. Open a CloudWatch datasource query in builder mode.
2. Select namespace `AWS/MediaTailor` and open the metric-name dropdown; it lists `Avails.*`.
3. With MediaTailor data in the account: the suggested `Avails.Duration` returns nothing, while manually entering `Avail.Duration` returns data.

**Root cause**
The hardcoded namespace-to-metrics map in `grafana-aws-sdk` (`pkg/cloudWatchConsts/metrics.go`, `AWS/MediaTailor` entry) uses the `Avails.` prefix. `pkg/tsdb/cloudwatch/services/hardcoded_metrics.go` serves suggestions from that map. The same map also feeds the Grafana Cloud CloudWatch scrape-job metric list and the generated CloudWatch services docs, so those need a dependency bump once the SDK is fixed.

**Additional context**
While verifying against the AWS docs, the `AWS/MediaTailor` list was also found to be missing several current metrics (`Avail.Impression`, `Avail.Complete`, the `Avail.Observed*` set, `GetManifest.Age`, `Requests`, `SkippedReason.*`); can be handled as a follow-up.

Reported via support.

lshippy:
Renamed metrics from 'Avails' to 'Avail' to match metric names listed here: https://docs.aws.amazon.com/mediatailor/latest/ug/monitoring-cloudwatch-metrics.html#metrics

Fixes: https://github.com/grafana/grafana/issues/130115


akkikumar72 (approved):
Thanks for the fix @lshippy 

Verified against the [AWS MediaTailor CloudWatch docs](https://docs.aws.amazon.com/mediatailor/latest/ug/monitoring-cloudwatch-metrics.html): the published names use the singular `Avail.` prefix (`Avail.Duration`, `Avail.FillRate`, `Avail.FilledDuration`), so this rename is correct. Also confirmed these three were the only `Avails.*` occurrences in the map.

